PR #5, opened 2023-10-11
Sync

ccruzagralopes:
- Sync Fall2023 release
- Configure python_tests as submodule


.version_information
```diff
@@ -1 +1 @@
-TH Fall2023 - beta3
```
mikaelhm, line 1: Should we leave the `.version_information` as `main - dev` on the `main` branch and only add a real version on the release branches?

mikaelhm (approved):

.version_information
ccruzagralopes: I think that's a good idea

mikaelhm: Up to you if want to that now, or we do it in a separate pr.

rquidute (approved):

PR #9, opened 2023-10-23
Add validation when getting a content line from uploaded file

rquidute:
### What changed

Add validation when getting a content line from uploaded file in `app/test_engine/models/manual_test_case.py`.
This is necessary because the uploaded file may contain char that are not decodable to utf-8.

### Related Issue
https://github.com/project-chip/certification-tool/issues/49

### Testing
Unit test 100% passed
Screenshot 2023-10-23 at 15 14 44

### DIFF
--- a/app/test_engine/models/manual_test_case.py
+++ b/app/test_engine/models/manual_test_case.py
@@ -153,10 +153,11 @@ class ManualLogUploadStep(TestStep, UserPromptSupport):
         logger.info("---- Start of Manual Log ----")
         with file.file as f:
             for line in f:
-                if file.content_type == "application/octet-stream":
+                try:
                     logger.info(line.decode("utf-8").strip())
-                else:
-                    logger.info(line.strip())
+                except UnicodeDecodeError:
+                    logger.warning("WARNING: The following line contained invalid UTF-8. Some content was replaced with: �")
+                    logger.info(line.decode("utf-8", errors='replace').strip())
         logger.info("---- End of Manual Log ----")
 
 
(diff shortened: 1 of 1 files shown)


antonio-amjr (approved):

ccruzagralopes (approved):

app/test_engine/models/manual_test_case.py
```diff
@@ -153,7 +153,10 @@ def handle_uploaded_file(self, file: UploadFile) -> None:
         logger.info("---- Start of Manual Log ----")
         with file.file as f:
             for line in f:
-                logger.info(line.decode("utf-8").strip())
+                if file.content_type == "application/octet-stream":
+                    logger.info(line.decode("utf-8").strip())
+                else:
+                    logger.info(line.strip())
```
mikaelhm, line 159: As discussed offline `line` here is always bytes, and we should always `decode` it as `utf-8`. This is not related to the `content_type` of the uploaded log.

The real problem here is that some `line` could contain invalid UTF-8 sequences. 

We have a couple options on how to handle it:
1. Simply reject any file that contains non-UTF-8 characters. 
2. Gracefully attempt to handle the places that this occur.

This proposal is one way to do it gracefully.

```suggestion
                try:
                    logger.info(line.decode("utf-8").strip())
                except UnicodeDecodeError:
                    logger.warning("WARNING: The following line contained invalid UTF-8. Some content was replaced with: �")
                    logger.info(line.decode("utf-8", errors='replace').strip())
```

mikaelhm (changes requested):
I'm not sure the proposed solution will work, I suggested an alternative by gracefully trying to decode the invalid UTF-8 by replacing invalid UTF-8.

Also, it would be great to have a test case for this in the repo with invalid UTF-8, to verify that this works.

app/test_engine/models/manual_test_case.py
rquidute: I agreed

mikaelhm (approved):

hiltonlima (approved):

PR #14, opened 2023-11-01
[Feature] Added Custom YAMLs

hiltonlima:
- Added folder for custom YAMLs files
- Added custom YAMLs collection 

-- With NO custom YAMLs...
image

-- With SOME custom YAMLs...
image

-- With 3 YAMLs. (one of each suite)...
image
image
image

PR #10, opened 2023-10-27
Refactoring test_collections folder

rquidute:
### What changed
This PR is the phase-1 for the feature **Python tests in TH UI**. 
The motivation to refactoring is to prepare the code to retrieve the python tests from SDK. 
The final proposal for the folders organization is described bellow (PS: **python_testing** folder is part of phase 2):

+ test_collections
    + sample_tests
    + python_tests
    + sdk_tests	
      + sdk_checkout
         + .version
         + yaml_tests
            + yaml
        + python_testing
           + script
        + sdk_runner
      + support
        + yaml_tests
          + models
        + python_testing
           + models
        + __init__.py

### Related Issues
Python tests in TH UI - Phase 1 (Refactoring test_collection folder) - https://github.com/project-chip/certification-tool/issues/72

### Testing

-   Unit test updated and passing
Screenshot 2023-10-27 at 15 47 47

-   Test case execution successfully in TH  (Automated, Semi-Automated, Simulated, Manual)

PR #18, opened 2023-11-07
Fix problem with sdk_tests collection using hardcoded paths

mikaelhm:
Since #10 and #14 was merged, we introduced a problem where the SDK tests only worked when running the backend in developer mode.

Specially, the problem stems from the fact that we used `/app/backed` hardcoded paths:
https://github.com/project-chip/certification-tool-backend/pull/10/files#diff-aea3c8e66d486bcaf287e7c19fa77772d3f40ac9733ee99c465657afda4104d5L41

And that only works in dev mode.

This PR changes this back to using relative paths.

### DIFF
--- a/test_collections/sdk_tests/support/paths.py
+++ b/test_collections/sdk_tests/support/paths.py
@@ -1,0 +1,21 @@
+#
+# Copyright (c) 2023 Project CHIP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


fabiowmm (approved):

raju-apple (approved):

gvargas-csa:
Screenshot 2023-11-07 at 7 16 50 p m

After updating to the new changes I ran into the problem that the test cases do not appear in production mode so I moved to this branch to verify it and it **works correctly**.

Screenshot 2023-11-07 at 8 14 15 p m

What I did to test it was add the fork branch to the current backend (`upstream`) project and test the changes from the fork (`mikaelhm:fix/hardcoded-paths`).

Screenshot 2023-11-07 at 8 00 02 p m

gvargas-csa (approved):

PR #19, opened 2023-11-08
Add python tests model classes

rquidute:
### What changed
This PR is the phase-3 for the feature **Python tests in TH UI**. 
This goal of this PR is to add the model classes related to Python Tests.

We need feedback if we need to move toward having a `common` model classes and `yaml` and `python` model classes inherit from that or we may keep using the way this PR is implemented by having yaml and python tests having it own attributes and methods.

### Related Issue
Python tests in TH UI - Phase 3 (Add Python Test model classes) - https://github.com/project-chip/certification-tool/issues/74

### Testing
-  Unit test added and passing
Screenshot 2023-11-08 at 13 54 45

- YAML test execution keep running successfully
Screenshot 2023-11-08 at 14 04 10

-  New Python Test Collection `SDK Python Test` is displayed in UI  (For development purpose I changed the TC_ACE_1_3 by adding the required methods in other to display this test case in UI)
Screenshot 2023-11-08 at 09 55 25

- Python Test steps are displayed in UI (They are shown as failed in the screenshot since the runner is under implementation, the goal of this PR is just to present the Test Case and list the steps.)

Screenshot 2023-11-08 at 14 12 26


mikaelhm:
I just tried to close/reopen to get the GitHub workflow to run unittests

mikaelhm:
If you rebase on `main` I think your PR should pass the checks.

test_collections/sdk_tests/support/python_testing/models/python_test_models.py
fabiowmm, line 1: Im my view the properties of these models are not YAML- or Python-specific, they are rather common to any TH test. I would move them up to a `support/models` level and keep a single copy of them using `TH` as a prefix instead of `Python` or `YAML`.

The fact that some property values may never be applicable to a Python test is not a problem. Maybe this same idea could be applied at other levels like TestCases and TestSuites and reduce code duplication without the need for subclassing...

fabiowmm (changes requested):

test_collections/sdk_tests/support/python_testing/sdk_python_test_tests.py
```diff
@@ -0,0 +1,110 @@
...
+from .models.python_test_parser import PythonParserException, parse_python_test
+from .models.test_declarations import (
+    PythonCaseDeclaration,
+    PythonCollectionDeclaration,
+    PythonSuiteDeclaration,
+)
+from .models.test_suite import SuiteType
+
+###
+# This file hosts logic load and parse Python test-cases, located in
```
ccruzagralopes, line 30:
```suggestion
# This file hosts logic to load and parse Python test cases, located in
```

test_collections/sdk_tests/support/python_testing/sdk_python_test_tests.py
```diff
@@ -0,0 +1,110 @@
...
+# The `sdk` sub-folder here is automatically maintained using the
+# `test_collections/sdk_tests/fetch_sdk_tests_and_runner.sh` script.
+#
+# The Python Tests are organized into 1 Test Suite:
+#        - Automated
+###
+
+SDK_PYTHON_TEST_PATH = Path(
+    "/app/backend/test_collections/sdk_tests/sdk_checkout/python_testing/scripts/sdk"
+)
```
ccruzagralopes, line 41: Please use the newly created `SDK_CHECKOUT_PATH` from `test_collections.sdk_tests.support.paths`

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def __handle_logs(self, logs: Any) -> None:
+        for log_entry in logs or []:
+            if not isinstance(log_entry, MatterLog):
+                continue
+
+            test_engine_logger.log(
+                CHIPTOOL_LEVEL,
+                CHIP_LOG_FORMAT.format(log_entry.module, log_entry.message),
+            )
```
ccruzagralopes, line 122: `logs` is defined as `str` in the methods that call this, not a list. How would this work?

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -0,0 +1,281 @@
...
+            if log_entry.is_error():
+                # Check if the step error came from exception or not, since the message
+                # in exception object has more details
+                # TODO: There is an issue raised in SDK runner in order to improve the
+                # message from log_entry:
+                # https://github.com/project-chip/connectedhomeip/issues/28101
+                if log_entry.exception:
+                    self.current_test_step.append_failure(log_entry.exception.message)
+                else:
+                    self.current_test_step.append_failure(log_entry.message)
```
ccruzagralopes, line 148: Is this relevant for python tests? Or does it only make sense for chip-tool/app test cases that use the websocket runner? Do the python tests implement this `logger`?

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -116,14 +226,56 @@ def __class_name(identifier: str) -> str:
...
+
 
 class PythonChipToolTestCase(PythonTestCase, ChipToolTest):
-    """Automated Python test cases."""
+    """Automated Python test cases using chip-tool."""
 
     test_type = ChipToolTestType.PYTHON_TEST
 
     def create_test_steps(self) -> None:
-        self.test_steps = [TestStep("Start Python test")]
```
ccruzagralopes: Should this really be removed? I talked about this in another comment already and why I believe it should be kept

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -0,0 +1,281 @@
...
+                            )
+                        if not callable(func):
+                            raise TypeError(f"{func_name} is not callable")
+                        # Call the method with the unpacked keyword arguments.
+                        func(**kwargs)
+
+                    for func_name, kwargs in update.items():
+                        call_function(self, func_name, kwargs)
+
+                handle_update(update)
```
ccruzagralopes, line 264: Could this `call_function` be defined somewhere else? I think it would be more readable if all this was reduced to just:

```python
for func_name, kwargs in update.items():
    call_function(self, func_name, kwargs)
```

...and the `call_function` method be defined outside of `execute`.

test_collections/sdk_tests/__init__.py
```diff
@@ -13,4 +13,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from .support.python_testing import sdk_python_collection
```
mikaelhm, line 16: do we also need a custom_python_collection?

test_collections/sdk_tests/support/models/th_test_models.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Any, Optional
+
+from pydantic import BaseModel
+
+###
+# This file declares YAML models that are used to parse the YAML Test Cases.
+###
+
+
+class THTestType(Enum):
```
mikaelhm, line 27: What does `TH` mean here? 

This seems to be the YAML models generalized? But does all this apply to the python tests?

test_collections/sdk_tests/support/python_testing/models/python_test_folder.py
```diff
@@ -0,0 +1,58 @@
...
+#
+from pathlib import Path
+
+from test_collections.sdk_tests.support.paths import SDK_CHECKOUT_PATH
+
+UNKNOWN_version = "Unknown"
+VERSION_FILE_FILENAME = ".version"
+
+
+class PythonTestFolder:
```
mikaelhm, line 24: This class looks almost like an exact copy of YamlTestFolder, except the file `python_file_paths` use `.py` instead of `.yaml`

Could we reuse this?

test_collections/sdk_tests/support/python_testing/models/python_test_models.py
```diff
@@ -0,0 +1,28 @@
...
+
+###
+# This file declares Python test models that are used to parse the Python Test Cases.
+###
+
+
+class PythonTest(THTest):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.type: THTestType.MANUAL  # type: ignore
```
mikaelhm, line 28: Why do we use type `MANUAL` aren't some of these tests fully automated?

test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -0,0 +1,126 @@
...
+        name=tc_info.desc, steps=tc_info.steps, config=tc_info.config, PICS=tc_info.pics
+    )
+    test.path = path
+    test.type = tc_info.type
+
+    return test
+
+
+def __extract_tcs_info(path: Path) -> PythonTestInfo:
+    # Currently PICS and config is not configured in Python Testing
```
mikaelhm, line 80: I think we need to figure out how to support PICS for these tests cases.

mikaelhm (changes requested):
Only a partial review. Will return to finish reviewing.

test_collections/sdk_tests/support/models/th_test_models.py
rquidute: @mikaelhm The ideia www to have a common model for both YAML and Python. Maybe we should use  `SDK` instead of `TH` or even remove `TH`. What do you think?

test_collections/sdk_tests/support/python_testing/models/python_test_folder.py
rquidute: Make sense. I'll do this.

test_collections/sdk_tests/support/python_testing/models/python_test_models.py
rquidute: Done.

mikaelhm (changes requested):
There's some good stuff here, but I think it requires some work still.

Regarding running the python tests, I suggest that we mount the actual python scripts fetched from the SDK in the SDK/chip-tool container so we know we're running the same script we've parsed for information.

test_collections/sdk_tests/support/python_testing/models/test_declarations.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, test: PythonTest, python_test_version: str) -> None:
+        super().__init__(
+            PythonTestCase.class_factory(
+                test=test, python_test_version=python_test_version
+            )
+        )
+
+    @property
+    def test_type(self) -> THTestType:
+        return self.class_ref.python_test.type
```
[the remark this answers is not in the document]
rquidute: I agree, but can we do this in a new PR, since this PR is already big?

test_collections/sdk_tests/__init__.py
rquidute: I think it might make sense.I filled an issue for that: https://github.com/project-chip/certification-tool/issues/97

test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -0,0 +1,135 @@
...
+BODY_INDEX = 0
+
+
+class PythonParserException(Exception):
+    """Raised when an error occurs during the parser of python file."""
+
+
+class PythonTestInfo:
+    """This class stores all the information from a python test case that came from
+    python test script file."""
```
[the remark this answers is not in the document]
rquidute: Agreed

test_collections/sdk_tests/support/python_testing/models/test_declarations.py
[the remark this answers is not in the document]
fabiowmm: I think this is true not only for this part of the code but for a lot more. I suspect that the only code that needs to be specialized is the parsing itself, everything else could be generalized for both YAML and Python tests.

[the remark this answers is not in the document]
rquidute: If we proceed with the plan to do this in a new PR, there is an issue for that: Python tests in TH UI - Phase 4 (code refactoring) -  https://github.com/project-chip/certification-tool/issues/75

test_collections/sdk_tests/support/yaml_tests/models/yaml_test_models.py
```diff
@@ -13,38 +13,18 @@
...
-    name: str
-    PICS: set[str] = set()
-    config: dict[str, Any]
-    steps: list[YamlTestStep] = Field(alias="tests")
-    type: YamlTestType = YamlTestType.MANUAL
-    path: Optional[Path]
+class YamlTest(YamlModelMixin, MatterTest):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(steps=kwargs["tests"], **kwargs)
+        self.type: MatterTestType.MANUAL  # type: ignore
```
antonio-amjr, line 30: Did you mean:

```suggestion
        self.type: MatterTestType = MatterTestType.MANUAL  # type: ignore
```
**or**
```suggestion
        self.type = MatterTestType.MANUAL  # type: ignore
```

rquidute: Thanks @antonio-amjr you are right.

antonio-amjr: Also, there is another one like in the file:
`python_test_models.py`

test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -0,0 +1,119 @@
...
+    """Raised when an error occurs during the parser of python file."""
+
+
+def parse_python_test(path: Path) -> PythonTest:
+    """Parse a single Python test file into PythonTest model.
+
+    This will also annotate parsed python test with it's path and test type.
+    """
+
+    return __parse_test_case_from_file(path)
```
ccruzagralopes, line 42: Why not make `__parse_test_case_from_file` public and use it directly? Or maybe keep this function name and use `__parse_test_case_from_file`'s body

test_collections/sdk_tests/support/python_testing/models/rpc_client/test_harness_client.py
```diff
@@ -0,0 +1,71 @@
...
+    class TestRunnerHooks:
+        pass
+
+
+MATTER_DEVELOPMENT_PAA_ROOT_CERTS = "/paa-root-certs"
+
+# Pre-computed param list for each Python Test as defined in Verification Steps.
+test_params = {
+    "TC_ACE_1_3": MatterTestConfig(
+        tests=["test_TC_ACE_1_3"],
```
antonio-amjr, line 37: Maybe I'm outdated here, but I remember omitting this `tests=[]` parameter.
Passing the "Test Class" will result running all the tests from the file if `tests` doesn't exist.
And that would keeps this simpler.

Can you try it out @rquidute?

test_collections/sdk_tests/support/python_testing/sdk_python_test_tests.py
antonio-amjr, line 1: Strange file naming for me this one. With `_test_tests`.
Could we change to `sdk_python_tests.py` or similar?

test_collections/sdk_tests/support/python_testing/models/rpc_client/test_harness_client.py
```diff
@@ -0,0 +1,71 @@
...
+        commissioning_method="on-network",
+        discriminators=[3840],
+        setup_passcodes=[20202021],
+        dut_node_ids=[0x12344321],
+        paa_trust_store_path=MATTER_DEVELOPMENT_PAA_ROOT_CERTS,
+        storage_path="/root/admin_storage.json",
+    )
+}
```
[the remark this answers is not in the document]
fabiowmm: +1

test_collections/sdk_tests/support/python_testing/models/rpc_client/test_harness_client.py
antonio-amjr: Exactly. It should be one test per file.
That's why I'm saying we should omit the `tests=[]` parameter.
Because the test file that should have only one entry point to test.

cecille: That's not a good assumption - we do have files where there is more than one test_ per file and this is intention because they share a setup

cecille: The default main scrapes the tests_ by introspection

antonio-amjr: Cecille, can you give more insights about this?

So you're saying we need the `tests=[]` parameter?
If there are python tests with more than one `test_` method, we would have to call more than one?

antonio-amjr: Using the `parse_matter_test_args()` it's great and simpler anyway.
This was changed so closing up.

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -0,0 +1,197 @@
...
+        for func_name, kwargs in update.items():
+            self.__call_function_from_name(self, func_name, kwargs)
+
+    def __call_function_from_name(self, obj, func_name, kwargs) -> None:  # type: ignore
+        func = getattr(obj, func_name, None)
+        if not func:
+            raise AttributeError(f"{func_name} is not a method of {obj}")
+        if not callable(func):
+            raise TypeError(f"{func_name} is not callable")
+        func(**kwargs)
```
mikaelhm, line 191: I think all of this stuff is drowning a bit in here, and I would like to see if we could refactor this and move it to a separate class that takes the TestCase instance and modifies it directly instead of the `__call_function_from_name`

I think @gladystonfranca mentioned offline there was some challenges around threads and asyncio execution loops, but I would like to see if we can clean this up, so it's simpler.

test_collections/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -0,0 +1,197 @@
...
+        class_name = cls.__class_name(identifier)
+        title = identifier
+
+        return type(
+            class_name,
+            (cls,),
+            {
+                "python_test": test,
+                "python_test_version": python_test_version,
+                "chip_tool_test_identifier": class_name,
```
mikaelhm, line 110: Do we need this? `chip_tool_test_identifier`

mikaelhm (changes requested):

app/chip_tool/chip_tool.py
```diff
@@ -661,32 +687,37 @@ def __trace_file_params(self, topic: str) -> str:
...
 
-    def set_pics(self, pics: PICS) -> None:
-        """Sends command to chip tool to create pics file inside the container.
+    def set_pics(self, pics: PICS, in_container: bool) -> None:
+        """Sends command to create pics file.
 
         Args:
             pics (PICS): PICS that contains all the pics codes
+            in_container (bool): Whether the file should be created in the SDK container
+                                or not. YAML tests run directly in the backend and
```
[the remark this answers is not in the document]
ccruzagralopes: > Is a file format required for the python runner?

Yes, the runner's test parser expects a file path

ccruzagralopes (approved):

test_collections/sdk_tests/support/python_testing/models/python_testing_hooks_proxy.py
```diff
@@ -13,10 +13,91 @@
...
+
+class SDKPythonTestResultBase(BaseModel):
+    type: SDKPythonTestResultEnum
+
+    def params_dict(self) -> dict:
+        return self.dict(exclude={"type"})
+
+
+class SDKPythonTestResultStart(SDKPythonTestResultBase):
+    type: Literal[SDKPythonTestResultEnum.START]
```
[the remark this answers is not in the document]
mikaelhm: I think if you use the literal you don't need to pass the `type` in the initializer

antonio-amjr (approved):

mikaelhm (approved):

fabiowmm (approved):

app/chip_tool/chip_tool.py
```diff
@@ -661,32 +687,37 @@ def __trace_file_params(self, topic: str) -> str:
         path = Path(DOCKER_LOGS_PATH) / filename
         return f'--trace_file "{path}" --trace_decode 1'
 
-    def set_pics(self, pics: PICS) -> None:
-        """Sends command to chip tool to create pics file inside the container.
+    def set_pics(self, pics: PICS, in_container: bool) -> None:
```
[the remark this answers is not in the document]
rquidute: This will be changed in a new PR

PR #29, opened 2023-11-14
Fix decommissioning problem in simulated test cases

ccruzagralopes:
## What changed

- Updated SDK SHA to get the latest runner changes.
- The runner's `start` and `stop` methods are now called one during test suite `setup` and `cleanup`. `stop` is only called after asking the user to decommission.

## Related issue

- https://github.com/project-chip/certification-tool/issues/40

## Testing

Ran one of each SDK YAML test cases and they were all executed as expected. The decommissioning after the simulated test case execution worked.

Screenshot 2023-11-14 at 14 24 27

### DIFF
--- a/app/chip_tool/chip_tool.py
+++ b/app/chip_tool/chip_tool.py
@@ -322,9 +322,10 @@ class ChipTool(metaclass=Singleton):
         return exit_code
 
     async def stop_chip_tool_server(self) -> None:
-        await self.__test_harness_runner.start()
+        await self.start_runner()
         await self.__test_harness_runner._client.send("quit()")
         self.__wait_for_server_exit()
+        self.stop_runner()
         self.__server_started = False
 
     def __get_gateway_ip(self) -> str:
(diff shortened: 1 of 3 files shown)


app/chip_tool/chip_tool.py
```diff
@@ -322,10 +322,12 @@ def __wait_for_server_exit(self) -> Optional[int]:
...
-        await self.__test_harness_runner.start()
-        await self.__test_harness_runner._client.send("quit()")
-        self.__wait_for_server_exit()
-        self.__server_started = False
+        if self.__server_started:
+            await self.start_runner()
+            await self.__test_harness_runner._client.send("quit()")
+            self.__wait_for_server_exit()
+            await self.stop_runner()
+            self.__server_started = False
```
mikaelhm, line 330: I'm wondering if we need some refactoring of this `ChipTool` class (maybe not needed for this PR)

But it's getting quite bloated, and it's really hard to follow what exactly is happening in a method like this.

I think we should split it in three classes:
- ChipCertBinsContainer
- ChipTool (covering running chip-tool/chip-app)
- MatterSDKRunner (for running yaml tests)

mikaelhm: What do you think @ccruzagralopes

mikaelhm (approved):

rquidute (approved):

app/chip_tool/chip_tool.py
mikaelhm: @ccruzagralopes can you make a new issue for this? An maybe propose some organization of the content in these classes, my suggestion above might not be the best one, so if you have other suggestions lets consider that too.

ccruzagralopes: Issue created: https://github.com/project-chip/certification-tool/issues/95

fabiowmm (approved):

app/chip_tool/chip_tool.py
ccruzagralopes: Resolving this conversation because this is blocking the merge, is not related to changes meant to be done in this PR and Mikael is currently on holiday

PR #39, opened 2023-12-08
Run multiple python tests in one execution

ccruzagralopes:
## What changed

- Added support to run multiple python tests in one test run execution.
- This was done by commissioning the DUT on test suite setup and omitting the `commissioning-method` argument when running the test cases.
- Added the `commission` command to `test_harness_client.py`.
- Shut down the server after the test execution completes.
- Added support to parse python test files that don't follow the template.

## Testing

Screenshot 2023-12-01 at 15 11 12


test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -73,23 +73,31 @@ def __parse_test_case_from_class(
...
     methods = [m for m in class_.body if isinstance(m, ast.FunctionDef)]
+
+    tc_desc = tc_name
+    tc_steps = []
+    tc_pics = []
+
     try:
         desc_method = next(m for m in methods if desc_method_name in m.name)
         tc_desc = desc_method.body[BODY_INDEX].value.value  # type: ignore
+    except StopIteration:  # Raised when `next` doesn't find a matching method
```
mikaelhm, line 84: We can also just pass a default value of `None` to `next`.

eg. 

```py
desc_method = next(m for m in methods if desc_method_name in m.name, None)
if desc_method:
    tc_desc = desc_method.body[BODY_INDEX].value.value
```

mikaelhm: have we considered putting this in a helper?
 
 ```py
 def get_method_by_name(name:str, methods:list) -> Optional[method]:
     return next(m for m in methods if name in m.name, None)
 ```
 
 Not sure if it's helpful

mikaelhm (approved):
Just a minor comment. I think it's simpler to remove all he try except when finding methods in the python test

hiltonlima (approved):

antonio-amjr (approved):

PR #44, opened 2023-12-13
Removed checking for valid arguments for python tests execution

rquidute:
## What changed
- Removed the checking for valid arguments when running python tests. We intend to remove any dependency regarding python arguments between TH and SDK. If any parameter in Python test were added in SDK we would have to update the valid args list in TH backend code.

- Removed `endpoint_id` attribute in `default_test_environment.config` because this parameter is no longer defined in the pydantic model and must be defined in test_parameter when necessary.
## Testing

- The way the arguments is defined in Project Config it will be used while running Python tests.

Screenshot 2023-12-13 at 10 44 54

Screenshot 2023-12-13 at 11 11 01

- Unit Testing updated and running
Screenshot 2023-12-13 at 10 56 19

### DIFF
--- a/test_collections/sdk_tests/support/python_testing/models/utils.py
+++ b/test_collections/sdk_tests/support/python_testing/models/utils.py
@@ -22,7 +22,6 @@ import loguru
 
 from app.schemas.test_environment_config import TestEnvironmentConfig
 from app.test_engine.logger import PYTHON_TEST_LEVEL
-from app.test_engine.logger import test_engine_logger as logger
 
 # Command line params
 RUNNER_CLASS_PATH = "/root/python_testing/test_harness_client.py"
@@ -32,31 +31,6 @@ EXECUTABLE = "python3"
 def generate_command_arguments(
     config: TestEnvironmentConfig, omit_commissioning_method: bool = False
 ) -> list:
-    # All valid arguments for python test
-    valid_args = [
-        "ble_interface_id",
-        "commissioning_method",
-        "controller_node_id",
-        "discriminator",
-        "endpoint",
-        "logs_path",
-        "PICS",
-        "paa_trust_store_path",
-        "timeout",
-        "trace_to",
-        "int_arg",
-        "float_arg",
-        "string_arg",
-        "json_arg",
-        "hex_arg",
-        "bool_arg",
-        "storage_path",
-        "passcode",
-        "dut_node_id",
-        "qr_code",
-        "manual_code",
-    ]
-
     dut_config = config.dut_config
     test_parameters = config.test_parameters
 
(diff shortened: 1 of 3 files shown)


test_collections/sdk_tests/support/python_testing/models/utils.py
```diff
@@ -76,13 +50,8 @@ def generate_command_arguments(
...
     if test_parameters:
         for name, value in test_parameters.items():
-            if name in valid_args:
-                if str(value) != "":
-                    arguments.append(f"--{name.replace('_','-')} {str(value)}")
-                else:
-                    arguments.append(f"--{name.replace('_','-')} " "")
-            else:
-                logger.warning(f"Argument {name} is not valid")
+            arg_value = str(value) if str(value) != "" else " "
```
ccruzagralopes, line 53: Is this `" "` space needed? Why not just let the empty string?

test_collections/sdk_tests/support/tests/python_tests/test_utils.py
```diff
@@ -49,7 +49,7 @@ async def test_generate_command_arguments_on_network() -> None:
         "--passcode 1234",
         "--commissioning-method on-network",
         "--paa-trust-store-path /paa-root-certs",
-        "--storage-path /root/admin_storage.json",
+        "--storage_path /root/admin_storage.json",
```
hiltonlima, line 52: do we need to replace "-" with "_" here?

test_collections/sdk_tests/support/python_testing/models/utils.py
rquidute: I agree with you, I was keeping that way it was before, but I think it not required to be `" "`

test_collections/sdk_tests/support/python_testing/models/utils.py
```diff
@@ -76,13 +50,8 @@ def generate_command_arguments(
...
         for name, value in test_parameters.items():
-            if name in valid_args:
-                if str(value) != "":
-                    arguments.append(f"--{name.replace('_','-')} {str(value)}")
-                else:
-                    arguments.append(f"--{name.replace('_','-')} " "")
-            else:
-                logger.warning(f"Argument {name} is not valid")
+            arg_value = str(value) if str(value) != "" else ""
+            arguments.append(f"--{name} {str(arg_value)}")
```
ccruzagralopes, line 54:
```suggestion
            arguments.append(f"--{name} {str(value)}")
```

ccruzagralopes (approved):

test_collections/sdk_tests/support/tests/python_tests/test_utils.py
mikaelhm: maybe add a comment about that here @rquidute

rquidute: Done

hiltonlima (approved):

mikaelhm (approved):

PR #45, opened 2023-12-19
Support multi-test scripts

ccruzagralopes:
## What changed

- Modified the python tests parser to separate each test case from a single python script. They are now shown individually in the UI and can be run separately.

## Testing

- I modified a python script to follow the template and provide the test steps for the tests contained in the file.
Screenshot 2023-12-19 at 08 38 26

- I ran the test cases in debug mode and manually performed a factory reset between each test execution to be able to pass most of these tests (a fix for this will be developed soon).
Screenshot 2023-12-19 at 09 49 50

### DIFF
--- a/test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
+++ b/test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
@@ -58,13 +58,13 @@ def parse_python_script(path: Path) -> list[PythonTest]:
 
     test_classes = __test_classes(parsed_python_file)
 
+    test_cases: list[PythonTest] = []
     for c in test_classes:
         test_methods = __test_methods(c)
         test_names = __test_case_names(test_methods)
 
-        test_cases: list[PythonTest] = []
         for test_name in test_names:
-            test_cases.append(__parse_test_case(test_name, test_methods, path))
+            test_cases.append(__parse_test_case(test_name, test_methods, c.name, path))
 
     return test_cases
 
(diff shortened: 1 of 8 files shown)


test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -28,42 +29,113 @@
...
-    and also the following methods:
-     * desc_[class_name] - This method should return a str with the test description
-     * pics_[class_name] - This method should return a list with the PICS required
-        for the test case
-     * steps_[class_name] - This method should return a list with the step description
-     * test_[class_name] -  This method contains the test logic
+    This method will search the file for classes that inherit from MatterBaseTest and
+    then look for methods with the following patterns to extract the needed information:
+     * test_[test_name] - (required) This method contains the test logic
+     * desc_[test_name] - (optional) This method should return a string with the test
```
rquidute, line 46:
```suggestion
     * desc_[test_name] - (required) This method should return a string with the test
```
Are we assuming desc_ is required?

rquidute (approved):

test_collections/sdk_tests/support/python_testing/models/python_test_parser.py
ccruzagralopes: I was thinking of using that method to identify whether the test case is following the template or not. If a test case doesn't have this method it should not break the parsing, so I'm not sure if this comment should say it's required, since the parsing will work if the method doesn't exist for a test case.

antonio-amjr (approved):

hiltonlima (approved):

PR #60, opened 2024-01-22
Create collections on init files

ccruzagralopes:
## What changed

- Created test collection objects on collection folder init files.
- Renamed "python_tests" collection to "Onboarding Payload Tests".
- Updated code around test collection discovery to stop automatically creating a collection from a subfolder of "test_collections" and only look for `TestCollectionDeclaration` objects.

## Reason

- To facilitate the collections migration to the program folder.

## Testing

- Collection "Onboarding Payload Tests" was discovered correctly.
Screenshot 2024-01-22 at 15 02 02

## Warning

There are changes in the python_tests submodule. So, before this PR can be merged, the following one must be merged and then the submodule SHA must be updated:
- https://github.com/project-chip/matter-test-scripts/pull/108

### DIFF
--- a/app/test_engine/test_collection_discovery.py
+++ b/app/test_engine/test_collection_discovery.py
@@ -106,19 +106,17 @@ def discover_test_collections(
     Note that disabled_test_cases and disabled_collections can be used to disable both
     entire collections or individual test cases.
     """
+    disabled_collections = disabled_collections or []
     collections: Dict[str, TestCollectionDeclaration] = {}
 
     names = __test_collection_folder_names()
-
-    # Apply filter if needed
-    if disabled_collections:
-        names = [n for n in names if n not in disabled_collections]
 
     for name in names:
         # Don't add collection if it doesn't have any suites
         if found_collections := __find_test_collections(name, disabled_test_cases):
             for collection in found_collections:
-                collections[collection.name] = collection
+                if collection.name not in disabled_collections:
+                    collections[collection.name] = collection
 
     return collections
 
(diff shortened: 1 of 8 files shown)


hiltonlima (approved):

rquidute (approved):

app/test_engine/test_collection_discovery.py
```diff
@@ -106,19 +106,17 @@ def discover_test_collections(
     Note that disabled_test_cases and disabled_collections can be used to disable both
     entire collections or individual test cases.
     """
+    disabled_collections = disabled_collections or []
```
fabiowmm, line 109: Does this means `disabled_collections` shouldn't be an Optional?

antonio-amjr (approved):

app/test_engine/test_collection_discovery.py
ccruzagralopes: `disabled_collections` has a default value that is read form a file. For the general application run, the disabled test cases will be read from this file. But for unit tests, the method is called without passing this parameter, so the value is None.

fabiowmm: Not sure I understand. If you don't pass the parameter, the value will be defined by the `disabled_test_collections()` function, that currently may return `None`. Why not make that function return `[]` instead of `None` and get rid of the Optional and of this line?

gladystonfranca (approved):

app/test_engine/test_collection_discovery.py
ccruzagralopes: Yeah, you're right. I'll update it

PR #92, opened 2024-05-16
Added log line sepator

hiltonlima:
- Let's separate the test suites using a line `#### ... ####`
- Let's separate the test cases using a line `==== ... ====`
- Let's separate the test steps using a line `---- ... ----`

# UI:
Screenshot 2024-05-16 at 17 27 27
Screenshot 2024-05-16 at 17 27 33
Screenshot 2024-05-16 at 17 27 37

# Log File (with 2 test suits):
Screenshot 2024-05-16 at 17 27 51

# Log File (with 2 test cases):
image

### DIFF
--- a/app/test_engine/models/test_case.py
+++ b/app/test_engine/models/test_case.py
@@ -20,6 +20,7 @@ from app.models import Project, TestCaseExecution
 from app.models.test_enums import TestStateEnum
 from app.schemas.test_environment_config import TestEnvironmentConfig
 from app.test_engine.logger import test_engine_logger as logger
+from app.test_engine.models.utils import LogSeparator
 from app.test_engine.test_observable import TestObservable
 from app.test_engine.test_observer import Observer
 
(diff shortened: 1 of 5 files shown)


app/test_engine/models/test_case.py
```diff
@@ -279,6 +281,11 @@ def next_step(self) -> None:
         self.current_test_step_index += 1
         self.current_test_step.mark_as_executing()
 
+    def __print_log_separator(self) -> None:
+        logger.info(
+            "=========================================================================="
+        )
+
```
rquidute, line 288:
```suggestion
        logger.info("=" * 80)

```

Could you check if this works?

rquidute (approved):

antonio-amjr (approved):

app/test_engine/models/test_case.py
hiltonlima: Done :)

ccruzagralopes (approved):
Nice!
I have a suggestion: I'm not sure if we have a file for constants, but maybe have constants like "test_<suite/case/step>_log_separator" and "log_separator_char_count" set in one file.

ccruzagralopes (approved):

PR #101, opened 2024-06-19
[PICS_V2_SUPPORT]Db changes for certification mode and mandatory test collections and test suites

rquidute:
### What changed
Added new column (certification_mode) to represent the certification mode at testrunexecution entity
Added new column (mandatory) to represent if the suite is mandatory

### Related issue
https://github.com/project-chip/certification-tool/issues/290

### Testing
Unit test passing
Screenshot 2024-06-19 at 17 55 02


hiltonlima (approved):

alembic/versions/e2c185af1226_pics_v2_support.py
```diff
@@ -0,0 +1,39 @@
...
+        "testrunexecution",
+        sa.Column("certification_mode", sa.Boolean(), nullable=False, default=False),
+    )
+    op.add_column(
+        "testsuiteexecution",
+        sa.Column("mandatory", sa.Boolean(), nullable=False, default=False),
+    )
+
+    op.execute("UPDATE testrunexecution SET certification_mode = false")
+    op.execute("UPDATE testsuiteexecution SET mandatory = false")
```
ccruzagralopes, line 31: Is this necessary? I believe that the default value is populated automatically in the existing records. Otherwise, I think we'd get an error and it would be necessary to first create the columns as nullable, update the existing records and then make it not nullable.

ccruzagralopes (approved):

alembic/versions/e2c185af1226_pics_v2_support.py
rquidute, line 31:
```suggestion
```

PR #103, opened 2024-06-21
Do not use fixed dir

hiltonlima:
issue: https://github.com/project-chip/certification-tool/issues/289

Not tested yet


hiltonlima:
Unable to get relative path because execution is in backend container. Path needs to be relative to host (rasp)

PR #111, opened 2024-07-23
Cancel test execution if one of the mandatory test cases fail on certification mode

ccruzagralopes:
Part of the PICS v2 implementation.

- Cancel test execution if one of the mandatory test cases fail on certification mode.
- Display a prompt explaining this cancellation to the user.
- Ask the user if the DUT should be commissioned at the start of each mandatory test case.

Screenshot 2024-07-23 at 14 30 39

### DIFF
--- a/app/test_engine/models/test_run.py
+++ b/app/test_engine/models/test_run.py
@@ -125,9 +125,7 @@ class TestRun(TestObservable):
                 self.test_run_execution.certification_mode
                 and test_suite.mandatory
                 and any(
-                    tc
-                    for tc in test_suite.test_cases
-                    if tc.state != TestStateEnum.PASSED
+                    tc.state != TestStateEnum.PASSED for tc in test_suite.test_cases
                 )
             ):
                 print("Abort execution")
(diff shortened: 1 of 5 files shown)


antonio-amjr (approved):

gladystonfranca (approved):

app/test_engine/models/test_run.py
```diff
@@ -120,6 +122,32 @@ async def __run_handle_errors(self) -> None:
...
+                self.__cancel_remaining_tests()
+                self.cancel()
+                await self.__display_mandatory_test_failure_prompt()
+                break
+
+    async def __display_mandatory_test_failure_prompt(self) -> None:
+        prompt = (
+            "At least one of the mandatory test cases have failed during a "
+            "test run execution in certification mode.\nAs a consequence, the remaining"
+            " tests have been cancelled."
```
hiltonlima, line 143: Just a suggestion: 
`At least one of the required test cases failed while running in certification mode. As a consequence, the remaining tests were cancelled.”`

hiltonlima (approved):

app/test_engine/models/test_run.py
ccruzagralopes: I'll keep "mandatory" instead of required because that's the word used in the test collection, but I'll update the rest. Thanks for the suggestion

PR #126, opened 2024-08-07
Fix the spell fault of DEFAULT_OTBR_INTERFACE in otbr_start.sh

bauerschwan:
The DEFAULT_OTBR_INTERFACE is being wrongly used as DEAFULT_OTBR_INTERFACE， if need to change the OTBR interface, it can not work.

Same fix as in https://github.com/project-chip/certification-tool/pull/244 but this time on the correct location


hiltonlima:
This fix will be done in your other PR: https://github.com/project-chip/certification-tool-backend/pull/127

PR #133, opened 2024-08-23
[Fix] Command Arguments For Python Test in BLE-Thread And BLE-WiFi Modes

antonio-amjr:
Fixes: https://github.com/project-chip/certification-tool/issues/351
Fixes: https://github.com/project-chip/certification-tool/issues/359

## Description
New command arguments were added for Python Tests,  related to the Thread and WiFi pairing modes.

For _BLE-Wi-Fi_ mode, the following arguments are now retrieved from the configuration and forwarded to the SDK container:
- WiFi SSID (“`--wifi-ssid`”)
- WiFi Password (“`--wifi-passphrase`”)
Screenshot 2024-08-26 at 16 31 25

For _BLE-Thread_ mode, only the bellow argument was added:
- Operational Dataset Hexa
Screenshot 2024-08-26 at 16 29 45
decimal (“`--thread-dataset-hex`”)

### Thread Dataset Options
To retrieve the Thread dataset, the user has two options:
1. Prepare a Thread dongle and provide all the Thread network related information (such as network_interface, rcp_serial_path, dataset, etc.)
2. Configure a external Thread Network and fill the TH project’s thread configuration only with the Operational Dataset Hexadecimal from that network

The item 1 case will trigger a local auto-configured Open Thread network container using the provided Thread dongle and the configuration so TH may pair and use the _BLE-Thread_ mode with the DUT.
Screenshot 2024-08-26 at 16 16 53

As for item 2, the user is responsible to configure the Thread Network manually for then retrive e provide the dataset to the TH.
Screenshot 2024-08-26 at 16 17 01

### Extra
Also in this change, the start script for an external otbr container was updatted for the same default thread configuration of TH.
Finally, the related unit tests were updated and a new one was added.


gladystonfranca (approved):

test_collections/matter/scripts/OTBR/otbr_start.sh
```diff
@@ -49,10 +49,10 @@ print_script_step "Waiting 10 seconds to give the the docker container enough ti
 sleep 10
 
 BR_CHANNEL_HEX=$(printf '%02x' $BR_CHANNEL)
-BR_PANID="5b${BR_VARIANT}"
```
fabiowmm, line 52: I think we should have comments here explaining the choice of values for each of these parameters.

fabiowmm (changes requested):
Good work. Just add some context to the thread parameter values as requested.

test_collections/matter/scripts/OTBR/otbr_start.sh
antonio-amjr: I have no idea about the choice of those values in the TH
All I did was to change the "example" values in this file to match with the default values already in the TH config.

The idea is the user change this file anyway for the desired thread configuration they want. Thus the TH as well.

antonio-amjr: We spoke offline and agreed to add some comments to this section.
Added [here](https://github.com/project-chip/certification-tool-backend/pull/133/commits/6701fac193c2ba9f88c0a215080f95df736c8b18)

fabiowmm (approved):

rquidute (approved):

PR #143, opened 2024-09-11
[FIX] Force container removal at destroy

rquidute:
### What changed
Added force remove and after container be destroyed.

### Related issue
https://github.com/project-chip/certification-tool/issues/414

### Testing
Unit test passing, and the TC-DRLK-2.1  is now passing
Screenshot 2024-09-11 at 18 45 39

### DIFF
--- a/test_collections/matter/sdk_tests/support/chip/chip_server.py
+++ b/test_collections/matter/sdk_tests/support/chip/chip_server.py
@@ -183,7 +183,8 @@ class ChipServer(metaclass=Singleton):
         except Exception as e:
             # Issue: https://github.com/project-chip/certification-tool/issues/414
             self.logger.info(
-                f"Could not get exit code after pkill command {__server_full_command}."
+                "Could not get exit code after pkill command "
+                f"{self.__server_full_command}."
             )
             self.logger.debug(str(e))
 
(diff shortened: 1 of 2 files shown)


test_collections/matter/sdk_tests/support/chip/chip_server.py
```diff
@@ -158,19 +158,6 @@ async def start(
...
-    def __wait_for_server_exit(self) -> Optional[int]:
-        if self.__chip_server_id is None:
-            self.logger.info(
-                "Server execution id not found, cannot wait for server exit."
-            )
-            return None
-
-        exit_code = self.sdk_container.exec_exit_code(self.__chip_server_id)
-        while exit_code is None:
-            exit_code = self.sdk_container.exec_exit_code(self.__chip_server_id)
```
hiltonlima, line 170: Instead of removing the wait completely... wouldn't it be interesting to add a timeout here?

That way, we would at least wait for the timeout. Since this problem is specific to a test case.

I believe it will have fewer side effects :)

ccruzagralopes (approved):

antonio-amjr (approved):

gladystonfranca (approved):

hiltonlima (approved):

PR #157, opened 2024-10-21
[FIX] Intermittent Tests Start Error With No Information

antonio-amjr:
Fixes: https://github.com/project-chip/certification-tool/issues/450

## Description
There's an intermittent problem when running many tests that the logs don't show any information on why the error occurred, as we can see in the image below:
Screenshot 2024-10-21 at 15 41 08

After trying to reproduce while debugging, the error was isolated and points out to be Python's `BaseManager` that failed to connect since the address was already used:
Screenshot 2024-10-15 at 16 33 27

## Solution
It was noted afterwards that the TH was currently creating the `BaseManager` (object to share data between processes) every time a Python Test is started. Thus, resulting sometimes in this error when was creating a new manager.

The proposed solution moved the manager creation to the initialization of the SDK Container, occurring only once for test run. When a Python Test starts now it will only access the manager previously created within the container.

> [!Note]
> Since it's an intermittent problem there's always a chance of coincidence, but this solution was anyway extensively tested for a couple of days and the problem didn't happen again.

### DIFF
--- a/test_collections/matter/sdk_tests/support/python_testing/models/test_case.py
+++ b/test_collections/matter/sdk_tests/support/python_testing/models/test_case.py
@@ -17,7 +17,6 @@ import re
 from asyncio import sleep
 from enum import IntEnum
 from inspect import iscoroutinefunction
-from multiprocessing.managers import BaseManager
 from pathlib import Path
 from socket import SocketIO
 from typing import Any, Optional, Type, TypeVar
@@ -38,10 +37,7 @@ from ...pics import PICS_FILE_PATH
 from ...sdk_container import SDKContainer
 from ...utils import prompt_for_commissioning_mode
 from .python_test_models import PythonTest, PythonTestType
-from .python_testing_hooks_proxy import (
-    SDKPythonTestResultBase,
-    SDKPythonTestRunnerHooks,
-)
+from .python_testing_hooks_proxy import SDKPythonTestResultBase
 from .utils import (
     EXECUTABLE,
     RUNNER_CLASS_PATH,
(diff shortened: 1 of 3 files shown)


test_collections/matter/sdk_tests/support/sdk_container.py
```diff
@@ -176,6 +179,10 @@ async def start(self) -> None:
             self.image_tag, self.run_parameters
         )
 
+        BaseManager.register("TestRunnerHooks", SDKPythonTestRunnerHooks)
+        self.manager = BaseManager(address=("0.0.0.0", 50000), authkey=b"abc")
+        self.manager.start()
```
ccruzagralopes, line 184: Maybe it would be good to mock these `register` and `start` method calls on any unit tests that execute this method

ccruzagralopes (approved):

test_collections/matter/sdk_tests/support/python_testing/models/test_case.py
```diff
@@ -254,10 +250,7 @@ async def execute(self) -> None:
         try:
             logger.info("Running Python Test: " + self.python_test.name)
 
-            BaseManager.register("TestRunnerHooks", SDKPythonTestRunnerHooks)
-            manager = BaseManager(address=("0.0.0.0", 50000), authkey=b"abc")
-            manager.start()
-            test_runner_hooks = manager.TestRunnerHooks()  # type: ignore
+            test_runner_hooks = self.sdk_container.manager.TestRunnerHooks()  # type: ignore
```
rquidute, line 253:
```suggestion
            test_runner_hooks = self.sdk_container.manager.TestRunnerHooks()  # type: ignore # flake8: noqa E501
```
Could try this in order to ignore `Line too long (92 > 88 characters) (E501)`

rquidute (approved):

hiltonlima (approved):

test_collections/matter/sdk_tests/support/sdk_container.py
ccruzagralopes: Looks good!

PR #158, opened 2024-10-22
Feature/446 add checking when no pics is informed

rquidute:
### What Changed
Added checking for PICS informed before starting a test run

`Important: DO NOT MERGE - Wainting for frontend implementation.`
Frontend issue: https://github.com/project-chip/certification-tool/issues/469

### Related Issue
https://github.com/project-chip/certification-tool/issues/446

### Testing
Development testing: backend returns error code 422. Frontend still not handling.
Screenshot 2024-10-22 at 10 32 19

New unit tests passing
Screenshot 2024-10-22 at 10 35 21

### DIFF
--- a/app/tests/api/api_v1/test_test_run_executions.py
+++ b/app/tests/api/api_v1/test_test_run_executions.py
@@ -797,6 +797,25 @@ async def test_test_run_execution_start(async_client: AsyncClient, db: Session) 
 
 
 @pytest.mark.asyncio
+async def test_test_run_execution_start_no_pics(
+    async_client: AsyncClient, db: Session
+) -> None:
+    test_run_execution = create_test_run_execution_with_some_test_cases(db=db, pics={})
+
+    # First attempt to start test run
+    response = await async_client.post(
+        f"{settings.API_V1_STR}/test_run_executions/{test_run_execution.id}/start",
+    )
+
+    # Assert 200 OK and that test run data is returned
+    assert response.status_code == HTTPStatus.UNPROCESSABLE_ENTITY
+    content = response.json()
+    assert isinstance(content, dict)
+    assert "detail" in content.keys()
+    assert content["detail"] == "No PICS were informed."
+
+
+@pytest.mark.asyncio
 async def test_test_run_execution_busy(async_client: AsyncClient, db: Session) -> None:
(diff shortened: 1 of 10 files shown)


app/tests/api/api_v1/test_test_run_executions.py
```diff
@@ -796,6 +796,25 @@ async def test_test_run_execution_start(async_client: AsyncClient, db: Session)
...
+    async_client: AsyncClient, db: Session
+) -> None:
+    test_run_execution = create_test_run_execution_with_some_test_cases(db=db, pics={})
+
+    # First attempt to start test run
+    response = await async_client.post(
+        f"{settings.API_V1_STR}/test_run_executions/{test_run_execution.id}/start",
+    )
+
+    # Assert 200 OK and that test run data is returned
```
ccruzagralopes, line 810: I believe that this comment is incorrect, because the expected status code is not 200 OK and no test run data should be returned

app/tests/utils/test_run_execution.py
```diff
@@ -85,12 +87,15 @@ def create_random_test_run_execution_archived(
 
 
 def create_random_test_run_execution(
-    db: Session, selected_tests: Optional[TestSelection] = {}, **kwargs: Any
+    db: Session,
+    selected_tests: Optional[TestSelection] = {},
+    pics: Optional[PICS] = PICS(),
```
ccruzagralopes, line 92: Maybe if you add this default value you wouldn't even need to update the other methods in this file?
Unless you prefer to keep the default value as an empty PICS, which is valid as well.
```suggestion
    pics: Optional[PICS] = create_random_pics(),
```

ccruzagralopes (approved):

app/tests/utils/test_run_execution.py
rquidute: You are right @ccruzagralopes, I'd rather keep PICS() just to keep empty as default value.

antonio-amjr (approved):

hiltonlima (approved):

app/tests/api/api_v1/test_test_run_executions.py
rquidute, line 810:
```suggestion
    # Assert 422 UNPROCESSABLE_ENTITY and a detail error message
```

PR #202, opened 2025-03-10
[FIX] Removing Performance Collection From PICS Validation

antonio-amjr:
Removing the Performance collection from PICS applicable tests validation.

This solves the issue from this TH version that using a **PICS** file results in auto-selecting the Performance test in background, even this collection not being visible in the "select tests" UI.

**Before:**
Screenshot 2025-03-10 at 09 58 15

**After:**
Screenshot 2025-03-10 at 09 59 05

### DIFF
--- a/app/pics_applicable_test_cases.py
+++ b/app/pics_applicable_test_cases.py
@@ -23,6 +23,9 @@ from app.test_engine.models.test_declarations import (
     TestCollectionDeclaration,
 )
 from app.test_engine.test_script_manager import test_script_manager
+from test_collections.matter.sdk_tests.support.performance_tests.sdk_performance_tests import (  # noqa
+    STRESS_TEST_COLLECTION,
+)
 
 
 def applicable_test_cases_list(pics: PICS) -> PICSApplicableTestCases:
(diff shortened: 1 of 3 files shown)


rquidute (approved):

app/pics_applicable_test_cases.py
```diff
@@ -69,6 +72,11 @@ def __applicable_test_cases(
 ) -> list:
     applicable_tests: list = []
 
+    # The Performance Test Collection should not be considered for the PICS.
+    # NOTE: The second parameter for the dictionary's "pop" method is provided so we may
+    # prevent a conditional exception when the following key is not present.
+    test_collections.pop(STRESS_TEST_COLLECTION, None)
```
ccruzagralopes, line 78: Will this remove the stress test collection from `test_script_manager.test_collections`? If so, please make a copy of the dictionary in the method `applicable_test_cases_list` and then pass that copy as a parameter here.

ccruzagralopes (approved):

app/pics_applicable_test_cases.py
antonio-amjr: Good catch Carol. Let me confirm this.
I believe you're right

hiltonlima (approved):

app/pics_applicable_test_cases.py
antonio-amjr: Thanks for the heads-up Carol. Done in https://github.com/project-chip/certification-tool-backend/pull/202/commits/f0ea1b12f53e6bdfcbd625d5641f31ade5c60c06

antonio-amjr:
FYI: This solution prevents the regular tests include the Performance test to the execution.
But there are still the other way around problem.

When a PICS file was added to the project configuration, the Stress & Stability execution will include the applicable tests from the PICS file, resulting in error.

The issue [issue#554](https://github.com/project-chip/certification-tool/issues/554) was created to track and solve this in the future since the Stress & Stability is not a broadly known or used feature yet.

PR #203, opened 2025-03-12
Version v2.12+spring2025

rquidute:
The goal of this PR is to update the SDK SHA and version information  for  `v2.12+spring2025` version

### DIFF
--- a/.version_information
+++ b/.version_information
@@ -1,1 +1,1 @@
-v2.12-beta3+spring2025
+v2.12+spring2025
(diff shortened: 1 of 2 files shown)


.version_information
```diff
@@ -1 +1 @@
-v2.12-beta3+spring2025
+v2.12+spring2025
```
hiltonlima, line 1:
```suggestion
v2.12+spring2025-develop
```

hiltonlima (approved):

.version_information
rquidute: This is the final version

rquidute:
Closing since the branch v2.12+spring2025 should be release branch

PR #212, opened 2025-04-23
Use discriminator for nfc thread

rquidute:
### What changed
The goal of this PR is to revert the changes introduced in PR #185, which made the discriminator field optional for NFC thread pairing.

### Testing
Unit tests have been updated accordingly and are all passing.

### DIFF
--- a/test_collections/matter/sdk_tests/support/yaml_tests/models/chip_suite.py
+++ b/test_collections/matter/sdk_tests/support/yaml_tests/models/chip_suite.py
@@ -129,7 +129,7 @@ class ChipSuite(TestSuite, UserPromptSupport):
         elif (
             self.config_matter.dut_config.pairing_mode is DutPairingModeEnum.NFC_THREAD
         ):
-            pair_result = await self.__pair_with_dut_nfc_thread()
+            pair_result = await self.__pair_with_dut_ble_thread()
         elif (
             self.config_matter.dut_config.pairing_mode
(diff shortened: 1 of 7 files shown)


test_collections/matter/sdk_tests/support/yaml_tests/models/chip_suite.py
```diff
@@ -129,7 +129,7 @@ async def __pair_with_dut(self) -> None:
         elif (
             self.config_matter.dut_config.pairing_mode is DutPairingModeEnum.NFC_THREAD
         ):
-            pair_result = await self.__pair_with_dut_nfc_thread()
+            pair_result = await self.__pair_with_dut_ble_thread()
```
OlivierGre, line 132: Isn't it confusing to use __pair_with_dut_ble_thread() when doing nfc_thread pairing?

Maybe that we should keep __pair_with_dut_nfc_thread() even if it is identical to __pair_with_dut_ble_thread()

ccruzagralopes, line 132: Should a NFC thread pairing really be executed as a BLE thread pairing?

rquidute: Code updated in next commit

OlivierGre (approved):
Looks good! Thank you!

ccruzagralopes (approved):

PR #216, opened 2025-04-29
[FIX] Adding Timeout To Server Exit

antonio-amjr:
This change adds a fail-safe mechanism by including a timeout when stopping the chip server, after a test execution of YAML tests.

Currently the code uses a while loop that waits only to the exit code return for killing the chip server process in the SDK.
That may never happen, so this PRs included a timeout logic for the chip server exit.

### DIFF
--- a/test_collections/matter/sdk_tests/support/chip/chip_server.py
+++ b/test_collections/matter/sdk_tests/support/chip/chip_server.py
@@ -19,7 +19,7 @@ from datetime import datetime
 from enum import Enum
 from pathlib import Path
 from random import randrange
-from time import time
+from time import sleep, time
 from typing import Generator, Optional, Union, cast
 
 import loguru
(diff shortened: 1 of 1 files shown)


ccruzagralopes (approved):

test_collections/matter/sdk_tests/support/chip/chip_server.py
```diff
@@ -165,10 +172,17 @@ def __wait_for_server_exit(self) -> Optional[int]:
             )
             return None
 
+        # A given timeout in seconds is provided to wait for the chip server exit code
+        # In case the timeout is triggered, the process continues after logging
+        timeout = time() + CHIP_SERVER_EXIT_TIMEOUT
         exit_code = self.sdk_container.exec_exit_code(self.__chip_server_id)
-        while exit_code is None:
+
+        while time() <= timeout:
```
oxesoft, line 180: I think the right way would be
`while exit_code is None and time() <= timeout:`
Also, it looks to be CPU intensive for me. What about adding a sleep()? I know it was this way, but the missing sleep() can even be the root cause of the issue.

antonio-amjr: Yeah, you're are correct, my bad.
And about the CPU intensive I agree. I'll make it sleep every 5 seconds.

Thanks for the feedbacks

antonio-amjr: Done in https://github.com/project-chip/certification-tool-backend/pull/216/commits/da205bb3fc01327d109a24fca74667724a3020e2

rquidute (approved):

antonio-amjr:
The logs for the sleep and exception error are as follow:
Screenshot 2025-04-29 at 17 23 36

oxesoft (approved):

PR #253, opened 2025-08-21
Updated unit test to not modify .version_information file

rquidute:
### What changed
The unit tests were refactored to prevent them from modifying the .version_information file.

### Related issue 
https://github.com/project-chip/certification-tool/issues/720

### Testing
Run the full suite of unit tests and verify that the .version_information file remains unchanged.

### DIFF
--- a/app/tests/version/test_read_test_harness_backend_version.py
+++ b/app/tests/version/test_read_test_harness_backend_version.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import os
+import tempfile
 from pathlib import Path
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


app/tests/version/test_read_test_harness_backend_version.py
```diff
@@ -45,53 +29,75 @@ def test_read_test_harness_backend_version() -> None:
...
+
+        version_file_path.write_text(expected_version_value)
+        sha_file_path.write_text(expected_sha_value)
+
+        with mock.patch("app.version.VERSION_FILEPATH", version_file_path), mock.patch(
+            "app.version.SHA_FILEPATH", sha_file_path
+        ), mock.patch.object(
+            target=utils_db,
+            attribute="get_db_revision",
+            return_value=expected_db_revision,
```
ccruzagralopes, line 46: I suggest also using this mock on the other tests

ccruzagralopes (approved):
Nice! I just have one small suggestion

app/tests/version/test_read_test_harness_backend_version.py
rquidute: Done

PR #256, opened 2025-09-02
[Camera] Add signaling endpoints for WebRTC

chulspro:
This PR adds signaling WebSocket endpoints for webrtc controller and peer.
This is to facilitate python TC to talk to browser UI to manage webrtc connections.

- This is to blindly forward the messages from webrtc peer (browser UI) to webrtc controller (python controller) and vice versa
- If a webrtc peer is already connected, reject new connections to not mix up signaling.

### DIFF
--- a/app/api/api_v1/sockets/web_sockets.py
+++ b/app/api/api_v1/sockets/web_sockets.py
@@ -51,3 +51,25 @@ async def websocket_video_endpoint(websocket: WebSocket) -> None:
     connection = WebSocketConnection(websocket, WebSocketTypeEnum.VIDEO)
     await socket_connection_manager.connect(connection)
     await socket_connection_manager.relay_video_frames(connection)
+
+
+@router.websocket("/ws/webrtc/controller")
+async def websocket_webrtc_controller_endpoint(websocket: WebSocket) -> None:
+    socket_connection_manager = SocketConnectionManager()
+    connection = WebSocketConnection(websocket, WebSocketTypeEnum.WEBRTC_CONTROLLER)
+    try:
+        if await socket_connection_manager.connect(connection):
+            await socket_connection_manager.start_webrtc_signaling(connection)
+    except Exception:
+        socket_connection_manager.disconnect(connection)
+
+
+@router.websocket("/ws/webrtc/peer")
+async def websocket_webrtc_peer_endpoint(websocket: WebSocket) -> None:
+    socket_connection_manager = SocketConnectionManager()
+    connection = WebSocketConnection(websocket, WebSocketTypeEnum.WEBRTC_PEER)
+    try:
+        if await socket_connection_manager.connect(connection):
+            await socket_connection_manager.start_webrtc_signaling(connection)
+    except Exception:
+        socket_connection_manager.disconnect(connection)
--- a/app/socket_connection_manager.py
+++ b/app/socket_connection_manager.py
@@ -17,7 +17,7 @@ import asyncio
 import json
 import socket
 from json import JSONDecodeError
-from typing import Callable, Dict, List, Union
+from typing import Callable, Dict, List, Optional, Union
 
 import pydantic
 from fastapi import WebSocket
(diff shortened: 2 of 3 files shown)


app/socket_connection_manager.py
```diff
@@ -52,23 +52,68 @@
...
         logger.info(
             f'Websocket disconnected: "{connection.websocket}"'
             f' of type: "{connection.type}".'
         )
-        self.active_connections.remove(connection)
+
+        if connection.type == WebSocketTypeEnum.WEBRTC_CONTROLLER:
+            self._signaling_webrtc_controller = None
+        elif connection.type == WebSocketTypeEnum.WEBRTC_PEER:
+            self._signaling_webrtc_peer = None
```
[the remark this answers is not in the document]
chulspro: There will always be only one controller (python TC) instance connected at a time. And for the peer the usage is also to have only one connection accepted at a time so 1:1 signaling can happen.

chulspro:
@oxesoft , @antonio-amjr , @rquidute : Could you please review this PR for Two way talk feature on Test Harness? Thank you for your time.

app/api/api_v1/sockets/web_sockets.py
```diff
@@ -51,3 +51,25 @@ async def websocket_video_endpoint(websocket: WebSocket) -> None:
...
+
+@router.websocket("/ws/webrtc/controller")
+async def websocket_webrtc_controller_endpoint(websocket: WebSocket) -> None:
+    socket_connection_manager = SocketConnectionManager()
+    connection = WebSocketConnection(websocket, WebSocketTypeEnum.WEBRTC_CONTROLLER)
+    try:
+        if await socket_connection_manager.connect(connection):
+            await socket_connection_manager.start_webrtc_signaling(connection)
+    except Exception:
+        socket_connection_manager.disconnect(connection)
```
antonio-amjr, line 64: Similar to the above, would be good to expand the try/catch with something like:

```suggestion
    except websockets.exceptions.ConnectionClosedError as e:
        socket_connection_manager.disconnect(connection)
        logger.error(f"WebSocket connection closed with an error: {e.code} - {e.reason}")
    except Exception as e:
        logger.error(f"An unexpected error occurred during WebSocket communication: {e}")
```

**PS:** Same question to the `webrtc/peer` below.

app/socket_connection_manager.py
```diff
@@ -188,5 +236,56 @@ async def __notify_invalid_message(
...
+            )
+
+    async def _webrtc_signal_error(
+        self, dest_type: WebSocketTypeEnum, error_msg: str, msg: str
+    ) -> None:
+        try:
+            data = json.loads(msg)
+            data["error"] = error_msg
+        except JSONDecodeError:
+            return
```
antonio-amjr, line 277:
```suggestion
        except JSONDecodeError:
            logger.error("Failed to parse JSON message")
            return
```

antonio-amjr (commented):
Please take a look on some comments below regarding the try/catch exceptions and the logging from them. Let me know if those make sense.

app/socket_connection_manager.py
chulspro: Thank you for your reivew. Since we are not really concerned about the type of exception occurred here, since with any type of exception we need to disconnect so that any future connections won't be blocked expanding try/catch is not needed. I have added an error log with the exception. Thanks

app/api/api_v1/sockets/web_sockets.py
```diff
@@ -51,3 +51,25 @@ async def websocket_video_endpoint(websocket: WebSocket) -> None:
     connection = WebSocketConnection(websocket, WebSocketTypeEnum.VIDEO)
     await socket_connection_manager.connect(connection)
     await socket_connection_manager.relay_video_frames(connection)
```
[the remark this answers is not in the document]
antonio-amjr: I see, fair enough.
If the relay_video_frames handle them it's ok.

Thanks for the reply Charles.

PR #257, opened 2025-09-03
[Camera] Adds pop up for two way talk verification request

chulspro:
This PR adds two way talk verification prompt required to validate two way talk test cases.

### DIFF
--- a/.version_information
+++ b/.version_information
@@ -1,1 +1,1 @@
-v2.14-beta1.1+winter2025
+v2.14-beta2+fall2025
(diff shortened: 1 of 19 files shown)


antonio-amjr (approved):
LGTM. But both comments pointed by Gemini seems fair.
Please take a look into using a constant variable for the `120 timeout` value and the code duplication from the show_prompts methods.

Either way, approved.

oxesoft (approved):

PR #260, opened 2025-09-09
Updated fetch_sdk_tests_and_runner script to install chipyaml wheel

rquidute:
### What changed
The goal of this PR is to fix another uncovered scenario in the initial PR. The execution was working only for the 1st execution because prestart script was not installing the python module, so this PR make the chipyaml module an installable package.

### Related Issue
 https://github.com/project-chip/certification-tool/issues/734
 
 ### Testing 
 TH launches successfully not only in the 1st execution
```
 Successfully installed chipyaml-0.0.1 matter-idl-1.0.0 matter-yamltests-1.0.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
+ exit 0
Prestart Complete
```

### DIFF
--- a/test_collections/matter/sdk_tests/scripts/fetch_sdk_tests_and_runner.sh
+++ b/test_collections/matter/sdk_tests/scripts/fetch_sdk_tests_and_runner.sh
@@ -173,7 +173,7 @@ build-backend = 'setuptools.build_meta'
 
 [project]
 name = "chipyaml"
-version = "0.0.1"
+version = "0.1.0"
 authors = [{name = "Project CHIP Authors"}]
 description = "Matter chipyaml package with adapters"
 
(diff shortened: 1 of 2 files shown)


test_collections/matter/sdk_tests/scripts/fetch_sdk_tests_and_runner.sh
```diff
@@ -159,28 +158,39 @@ rm -rf ${EXTRACTION_ROOT}
...
+
+# Create pyproject.toml for chipyaml package at the root level
+cat > ${EXTRACTION_ROOT}/chipyaml_src/pyproject.toml << 'EOF'
+[build-system]
+requires = ['setuptools>=45', 'wheel']
+build-backend = 'setuptools.build_meta'
+
+[project]
+name = "chipyaml"
+version = "0.0.1"
```
antonio-amjr, line 176: Having Major and Minor versions as "zero" seems strange 
```suggestion
version = "0.1.0"
```

antonio-amjr (approved):
LGTM

PR #286, opened 2025-12-18
Add support to sort the test run execution list and retrieve all records without pagination

rquidute:
### What changed
  This PR adds sort ordering functionality and the possibility to retrieve all test run execution records.

  - Sort Order Parameter: Added optional sort_order parameter to the /test_run_executions endpoint with "asc" default for
  backwards compatibility
  - All Results Support: Implemented limit=0 logic to return all records without pagination when requested
  
  ### Related Issue
  https://github.com/project-chip/certification-tool/issues/823
  
  ### Testing 

- Unit tests added and Passing
- Several backend endpoint calls such as  
    - GET /api/v1/test_run_executions/project_id=1&archived=false&limit=10&sort_order=desc
    - GET /api/v1/test_run_executions/project_id=1&archived=false&skip=18&limit=10&sort_order=asc
    - GET /api/v1/test_run_executions/project_id=1&archived=false&skip=18&limit=0&sort_order=asc

### DIFF
--- a/app/crud/crud_test_run_execution.py
+++ b/app/crud/crud_test_run_execution.py
@@ -62,7 +62,7 @@ class CRUDTestRunExecution(
         archived: Optional[bool] = False,
         search_query: Optional[str] = None,
         order_by: Optional[str] = None,
-        sort_order: Optional[str] = "asc",
+        sort_order: str = "asc",
         skip: Optional[int] = 0,
         limit: Optional[int] = 100,
     ) -> Sequence[TestRunExecution]:
--- a/app/crud/crud_test_run_execution.py
+++ b/app/crud/crud_test_run_execution.py
@@ -62,9 +62,9 @@ class CRUDTestRunExecution(
         archived: Optional[bool] = False,
         search_query: Optional[str] = None,
         order_by: Optional[str] = None,
-        sort_order: str = "asc",
         skip: Optional[int] = 0,
         limit: Optional[int] = 100,
+        sort_order: str = "asc",
     ) -> Sequence[TestRunExecution]:
         query = self.select()
 
@@ -110,9 +110,9 @@ class CRUDTestRunExecution(
         archived: Optional[bool] = False,
         search_query: Optional[str] = None,
         order_by: Optional[str] = None,
-        sort_order: str = "asc",
         skip: Optional[int] = 0,
         limit: Optional[int] = 100,
+        sort_order: str = "asc",
     ) -> List[TestRunExecutionWithStats]:
         results = self.get_multi(
             db=db,
--- a/app/crud/crud_test_run_execution.py
+++ b/app/crud/crud_test_run_execution.py
@@ -110,7 +110,7 @@ class CRUDTestRunExecution(
         archived: Optional[bool] = False,
         search_query: Optional[str] = None,
         order_by: Optional[str] = None,
-        sort_order: Optional[str] = "asc",
+        sort_order: str = "asc",
         skip: Optional[int] = 0,
         limit: Optional[int] = 100,
     ) -> List[TestRunExecutionWithStats]:
--- a/app/crud/crud_test_run_execution.py
+++ b/app/crud/crud_test_run_execution.py
@@ -85,19 +85,11 @@ class CRUDTestRunExecution(
                 )
             )
 
-        if order_by is None:
-            # Default to ordering by id with specified sort order
-            if sort_order == "desc":
-                query = query.order_by(self.model.id.desc())
-            else:
-                query = query.order_by(self.model.id.asc())
+        column = self.model.id if order_by is None else getattr(self.model, order_by)
+        if sort_order == "desc":
+            query = query.order_by(column.desc())
         else:
-            # Apply sort_order to the specified order_by column
-            column = getattr(self.model, order_by)
-            if sort_order == "desc":
-                query = query.order_by(column.desc())
-            else:
-                query = query.order_by(column.asc())
+            query = query.order_by(column.asc())
 
         query = query.offset(skip)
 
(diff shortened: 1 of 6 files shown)


app/tests/api/api_v1/test_test_run_executions.py
```diff
@@ -939,6 +939,41 @@ def test_read_multiple_test_run_executions_with_search_query(
...
 
 
+def test_read_multiple_test_run_executions_with_limit_zero_returns_all(
+    client: TestClient, db: Session
+) -> None:
+    """Test that limit=0 returns all test run executions."""
+
+    # Create several test executions to ensure we have more than default limit
+    test_runs = []
+    for i in range(5):
```
ccruzagralopes: I suggest maybe using a bigger number, which is greater than the default value (100)

app/tests/api/api_v1/test_test_run_executions.py
```diff
@@ -939,6 +939,41 @@ def test_read_multiple_test_run_executions_with_search_query(
...
+    response = client.get(f"{settings.API_V1_STR}/test_run_executions?limit=0")
+    assert response.status_code == HTTPStatus.OK
+    content = response.json()
+    assert isinstance(content, list)
+
+    # Verify that all our created test runs are in the response
+    created_ids = {tr.id for tr in test_runs}
+    response_ids = {tr["id"] for tr in content}
+
+    # All created test runs should be present (and potentially more from other tests)
```
ccruzagralopes, line 966: Is it ok for there to be more test executions than the ones created on this test?

app/tests/crud/test_test_run_execution.py
```diff
@@ -809,3 +809,79 @@ def test_import_execution_success_without_test_config() -> None:
...
+    )
+
+    # Test descending order
+    test_run_executions_desc = crud.test_run_execution.get_multi_with_stats(
+        db, project_id=project.id, sort_order="desc"
+    )
+
+    # Verify we have all test runs
+    assert len(test_run_executions_asc) >= 3
+    assert len(test_run_executions_desc) >= 3
```
ccruzagralopes: Maybe move this verification to later, so you can check against `asc_our_runs` and `desc_our_runs` instead (and maybe check for the exact length instead of `>=`)

ccruzagralopes (approved):

test_collections/matter/sdk_tests/support/python_testing/models/python_test_parser.py
```diff
@@ -42,6 +43,67 @@
...
+            # Wait before retrying
+            time.sleep(delay)
+
+    # This should never be reached, but just in case
+    if last_exception:
+        raise last_exception
+    else:
+        raise RuntimeError(
+            f"Failed to load JSON from {path} after {max_retries} attempts"
+        )
```
antonio-amjr: Can we use `finally` here to this "just in case" logic?
I believe has a better readability.

Although I think this is not necessary.
Did you face a case where this conditional was reached?

antonio-amjr (changes requested):
Please consider the following.
Let me know if you disagree.

antonio-amjr:
@rquidute, Did you try this solution with a execution of different projects IDS and running from several different folders?

test_collections/matter/sdk_tests/support/python_testing/models/python_test_parser.py
rquidute: The try .. except is inside for statement, so maybe there will be more iteration before a finally. So I think we cold keep this way by now.

app/tests/api/api_v1/test_test_run_executions.py
rquidute: Done

app/tests/crud/test_test_run_execution.py
rquidute: Done

oxesoft (approved):

app/crud/crud_test_run_execution.py
```diff
@@ -85,11 +86,19 @@ def get_multi(
...
 
         if order_by is None:
-            query = query.order_by(self.model.id)
+            # Default to ordering by id with specified sort order
+            if sort_order == "desc":
+                query = query.order_by(self.model.id.desc())
+            else:
+                query = query.order_by(self.model.id.asc())
         else:
             query = query.order_by(order_by)
```
antonio-amjr: So, if I pass `order_by` parameter, I can't use the sort feature?
Is that a way to use both?

antonio-amjr (commented):
Please take a look at the below:

app/crud/crud_test_run_execution.py
rquidute: Good point, let me think about it

app/tests/api/api_v1/test_test_run_executions.py
rquidute: At first, I don't think this could be a problem

app/crud/crud_test_run_execution.py
```diff
@@ -85,11 +86,24 @@ def get_multi(
             )
 
         if order_by is None:
```
antonio-amjr: Is the suggestion below possible to substitute this **whole** conditional?
What do you think?

```suggestion
            column = self.model.id if order_by is None else getattr(self.model, order_by)
            if sort_order == "desc":
                query = query.order_by(column.desc())
            else:
                query = query.order_by(column.asc())
```

antonio-amjr (approved):
**Please**, take a look in the **suggestion** above 👆
Approving either way

app/crud/crud_test_run_execution.py
rquidute: Done

PR #328, opened 2026-06-02
Remove unused Performance Tests feature

rquidute:
### Summary
  Remove the unused Performance Tests feature from the backend. This feature was built but never adopted by users, leaving a significant amount of dead
  code in the codebase.
  
  
#### Motivation

  The Performance Tests feature was implemented but is not being used by users in production. Keeping it adds unnecessary maintenance burden, increases cognitive overhead when reading the codebase, and adds dead volume mounts to every SDK container startup. Removing it reduces the codebase by ~2,044 lines.

#### Impact
  - No functional impact for users — the feature was not in use
  - The POST /{id}/performance_summary API endpoint is removed; any client calling it will receive a 404
  - SDK containers will no longer mount the three stress test script volumes
  - TestScriptManager logic is simplified — the iteration/loop-commissioning special-case is gone
  
 ### Related Issue 
[ #1014](https://github.com/project-chip/certification-tool/issues/1014)
 
 ### Testing
All  unit tests runs successfully 
TH runs successfully


oxesoft (approved):

oxesoft:
Cherry-picked to [v2.16-develop](https://github.com/project-chip/certification-tool-backend/tree/v2.16-develop)

PR #333, opened 2026-06-25
Feature/1024 project logs endpoint

rquidute:
### What changed
Adds a new backend API endpoint, GET /api/v1/projects/{id}/logs, that lets a client download all test run execution logs for a given project as a single ZIP archive, plus a comprehensive test suite covering the new behavior.

#### Changes

- New endpoint: GET /api/v1/projects/{id}/logs

  - Added to app/api/api_v1/endpoints/projects.py.
  - Looks up the project (404 if it doesn't exist), then fetches all TestRunExecution records for that project
  (crud.test_run_execution.get_multi(..., limit=0)).
  - Builds an in-memory outer ZIP (BytesIO + ZipFile) containing one entry per execution:
    - Flat mode (default, grouped=false): each execution's log is converted to plain lines via
  log_utils.convert_execution_log_to_list(...) and written as a <execution.id>-<execution.title>.log text file.
    - Grouped mode (grouped=true): each execution's log is grouped by test-case state via
  log_utils.group_test_run_execution_logs(...), packed into an inner ZIP via log_utils.create_grouped_log_zip_file(...), and written as a <execution.id>-<execution.title>.zip (i.e., a zip-of-zips).
  - Returns a StreamingResponse with media_type="application/zip" and a Content-Disposition: attachment header whose filename is derived from the project name (sanitized to alphanumerics, -, and _) or falls back to project-{id}-logs.zip if the project has no name.
  - Response schema/docs annotated for OpenAPI (binary zip content, example Content-Disposition header) so the generated CLI client documents the endpoint correctly.
  - A project with no executions returns a valid, empty ZIP rather than erroring.

#### Tests

  - New file app/tests/api/api_v1/test_projects_logs.py with 7 test cases covering:
    - Flat download returns a zip with one .log per execution, correct headers.
    - Grouped download (?grouped=true) returns a zip containing inner .zip files.
    - Empty project returns an empty but valid zip.
    - Filename is derived from the project name.
    - 404 for a non-existent project ID.
    - Exact entry-naming conventions for both flat (<id>-<title>.log) and grouped (<id>-<title>.zip) modes.

#### Motivation

  Previously there was no way to bulk-download all logs for every test run execution belonging to a project in one request — users had to fetch logs execution-by-execution. This endpoint provides a single, convenient bulk-export path (e.g., for archival, sharing with support, or offline analysis), and reuses existing log_utils helpers already used elsewhere for per-execution log formatting/grouping.

#### Impact

  - Additive, non-breaking: this is a new route; no existing endpoints, schemas, or behaviors are modified.
  - New dependency on log_utils in projects.py (module already exists in the codebase; only the import surface changes here).
  - Response can be large for projects with many/large executions since the whole ZIP is built in memory (BytesIO) before streaming — acceptable for now but worth keeping in mind for very large projects.
  - CLI's auto-generated API client should be regenerated (cli/scripts/generate_client.sh) to pick up the new endpoint once this merges.
  
### Related Issue
  https://github.com/project-chip/certification-tool/issues/1024
  
 ### Testing
- Unit tests added and passing
- Developer testing are ongoing

### DIFF
--- a/app/api/api_v1/endpoints/projects.py
+++ b/app/api/api_v1/endpoints/projects.py
@@ -16,17 +16,19 @@
 import json
 import traceback
 from http import HTTPStatus
+from io import BytesIO
 from typing import List, Sequence, Union
+from zipfile import ZipFile
 
 from fastapi import APIRouter, Depends, File, HTTPException, Request, UploadFile
 from fastapi.encoders import jsonable_encoder
-from fastapi.responses import JSONResponse
+from fastapi.responses import JSONResponse, StreamingResponse
 from loguru import logger
 from pydantic import ValidationError, parse_obj_as
 from sqlalchemy.orm import Session
 from sqlalchemy.orm.attributes import flag_modified
 
-from app import crud, models, schemas
+from app import crud, log_utils, models, schemas
 from app.db.session import get_db
 from app.default_environment_config import default_environment_config
 from app.models.project import Project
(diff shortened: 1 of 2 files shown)


app/api/api_v1/endpoints/projects.py
```diff
@@ -460,3 +462,82 @@ def importproject_config(
...
+                    grouped_logs=grouped_logs
+                )
+                entry_name = f"{execution.id}-{execution.title}.zip"
+                outer_zip.writestr(entry_name, inner_zip_buffer.read())
+            else:
+                log_lines = log_utils.convert_execution_log_to_list(
+                    log=execution.log, json_entries=False
+                )
+                entry_name = f"{execution.id}-{execution.title}.log"
+                outer_zip.writestr(entry_name, "\n".join(log_lines))
```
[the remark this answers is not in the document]
rquidute: Done

oxesoft (approved):

app/api/api_v1/endpoints/projects.py
```diff
@@ -372,6 +374,18 @@ def __project(db: Session, id: int) -> Project:
...
+def __safe_filename_component(value: Union[str, None], fallback: str) -> str:
+    """Sanitize a value for safe use as part of a file or zip entry name.
+
+    Keeps only alphanumeric characters, hyphens and underscores, replacing
+    everything else (including path separators like "/" or "\\") with "_".
+    This avoids path traversal / Zip Slip issues when the value originates
+    from user-controlled data (e.g. project name or execution title).
+    """
+    source = value or fallback
+    return "".join(c if c.isalnum() or c in "-_" else "_" for c in source)
```
antonio-amjr: Regex seems to be a better solution here:
```suggestion
    return re.sub(r'[^a-zA-Z0-9_-]', '_', source)
```
The generator expression works fine, but regex with negate character (^) is the standard Python idiom for this cases. Also, since `re` module is implemented in C, it is generally much faster.

app/api/api_v1/endpoints/projects.py
```diff
@@ -372,6 +374,18 @@ def __project(db: Session, id: int) -> Project:
     return project
 
 
+def __safe_filename_component(value: Union[str, None], fallback: str) -> str:
```
antonio-amjr: Giving fallback a default value (like "" or "unnamed") ensures the function never accidentally processes a NoneType.
```suggestion
def __safe_filename_component(value: Union[str, None], fallback: str = "default") -> str:
```

app/api/api_v1/endpoints/projects.py
rquidute: Done

app/api/api_v1/endpoints/projects.py
rquidute: Done

antonio-amjr (approved):

PR #342, opened 2026-07-20
Enable SRP server on OTBR container startup

khodya:
#### Summary

Explicitly enable the OpenThread SRP server when the OTBR container starts, instead of relying on `RoutingManager`'s auto-enable heuristic.

##### Problem

- Both OTBR startup paths (`otbr_start.sh` and `ThreadBorderRouter.form_thread_topology()` in `otbr_manager.py`) form the Thread topology (`dataset ... commit active`, `ifconfig up`, `thread start`, `netdata register`) but never issue `srp server enable`.
- The `nrfconnect/otbr` image relies on OpenThread's SRP server auto-enable mode, which only flips the server on once `RoutingManager` finishes its initial policy evaluation on the infra interface. In this environment (Docker with `--network host` on a Raspberry Pi host), that evaluation does not reliably complete, so the SRP server stays disabled indefinitely.
- Consequence observed on [TC-CNET-4.12](https://github.com/project-chip/connectedhomeip/pull/73007) step 7 (switching the DUT to a second Thread network hosted by a second OTBR): the DUT successfully joins the new Thread partition, but since the SRP server on that OTBR is disabled, the DUT's SRP client never registers its `_matter._tcp` host, so nothing is published for the test harness's mDNS/SRP query. `wait_for_srp_update()` polls for 180s and times out with `TestFailure: Unable to find device`, even though Thread connectivity itself is healthy.
- `ot-ctl netdata show` on the affected OTBR container confirmed the DNS/SRP service entry (service data `0x5d`) was absent from Network Data until the server was enabled manually via `ot-ctl srp server enable`.

##### Solution

- Added `srp server enable` right after `netdata register` in both OTBR startup flows:
  - `test_collections/matter/scripts/OTBR/otbr_start.sh` (`BR_PARAMS` list, shell/manual path).
  - `test_collections/matter/sdk_tests/support/otbr_manager/otbr_manager.py` (`ThreadBorderRouter.form_thread_topology()`, harness-managed path).
- This makes SRP server activation deterministic at topology formation time instead of depending on auto-enable succeeding on the host's infra interface. The setting persists across later `dataset set active` / `thread start` calls made when switching Thread datasets mid-test, so it only needs to be set once per container lifetime.
- The repo already had a manual workaround script (`otbr_srp_restart.sh` doing `srp server disable` + `srp server enable`), but it was not wired into any startup path — this change makes the fix automatic.

#### Testing

- Manually reproduced the failure: ran `TC_CNET_4_12.py` against a two-OTBR Thread setup; step 7 failed with `Unable to find device` while `ot-ctl netdata show` on the second OTBR showed no `0x5d` DNS/SRP service entry.
- Confirmed `ot-ctl srp server enable` on the affected container immediately publishes the `0x5d` service entry in Network Data and allows the DUT's SRP registration to be discovered.
- No automated test added: this change only affects the local OTBR container bring-up scripts used by the certification harness/manual test setup, not a code path with its own unit tests.


cecille (approved):

rquidute (approved):

greens (approved):
Do we know why `RoutingManager`'s policy evaluation doesn't reliably complete?

khodya:
> Do we know why `RoutingManager`'s policy evaluation doesn't reliably complete?

No, this is something that requires investigation, probably by tracing the RoutingManager logs.

antonio-amjr (approved):

khodya:
I don't have access to tick the box to merge this PR.

khodya:
@Mergifyio queue

rquidute:
Hi @khodya, I changed the target branch to the one used in v1.7. Let me know if we can proceed with merging into this branch

rquidute:
Merged into [v2.16-develop](https://github.com/project-chip/certification-tool-backend/tree/v2.16-develop) branch (Matter v1.7 version)

PR #348, opened 2026-08-13
Improve OTBR startup/shutdown reliability and diagnostics (#1071)

rquidute:
## Summary

Related to project-chip/certification-tool#1071 ("Border router does not start properly" / "connect session failed" failures).

This does **not** claim to be a confirmed root-cause fix for the reporter's exact original failure — we don't yet have `docker logs`/container state from their box to confirm the mechanism. It addresses two real, reproducible issues found while investigating:

1. **No diagnostics on OTBR startup failure.** `otbr_start.sh` used a fixed 10s sleep after starting the container, then ran `ot-ctl` commands blind. If `otbr-agent` failed to come up, every subsequent `ot-ctl` call failed with an opaque `connect session failed: No such file or directory` and no indication of why.
2. **Orphaned OTBR containers can block the next start.** `otbr_manager.py`'s `ThreadBorderRouter` tracks its container only via an in-memory reference and creates it without an explicit name. If a previous `start_device()` call fails partway, or the backend process restarts, that reference is lost — but the container can still be running and holding the RCP serial device (Docker only checks `container.status == "running"`, not whether `otbr-agent` actually attached). This blocks the next `start_device()` call from acquiring the RCP, producing `Border router does not start properly for nrfconnect/otbr:9185bda`. Reproduced locally: an orphaned container survived a failed `start_device()` call and blocked a subsequent test run with this exact error until manually removed.

   The same blind spot exists on the manual path: `otbr_start.sh` only stops/removes a container literally named `otbr-chip`, so a TH-managed container left running under Docker's random name (e.g. from a prior test run) is invisible to it — switching from TH-managed OTBR to a manual `otbr_start.sh` run can hit `connect session failed` on the very first `ot-ctl` call because the RCP is still held by the orphan.

## Changes

**`otbr_start.sh`**
- Fail fast with a clear message if the RCP device (`/dev/ttyACM0`) is missing, instead of proceeding into a container that can't attach to anything.
- Replace the fixed 10s sleep with a readiness poll (`ot-ctl state`, 30s timeout).
- Dump `docker logs otbr-chip` on any `ot-ctl` failure (readiness or dataset setup) before exiting, so the actual failure is visible from a single log capture.
- In addition to removing a container named `otbr-chip`, also remove any other container running the OTBR image (covers a leftover TH-managed container from a previous run).

**`otbr_stop.sh`**
- Use `docker stop` (graceful) instead of `docker kill`, giving `otbr-agent` a chance to release the RCP serial device cleanly before the container exits.

**`app/container_manager/container_manager.py`**
- Add `remove_containers_for_image()`: force-removes any container (running or stopped) built from a given image.

**`test_collections/matter/sdk_tests/support/otbr_manager/otbr_manager.py`**
- Call `remove_containers_for_image()` in `start_device()` before creating a new OTBR container, so a leftover/untracked container from this image can't block a fresh start.

## Testing

- Manually verified `otbr_start.sh`/`otbr_stop.sh` syntax (`bash -n`) and reviewed the full diff.
- Reproduced the orphaned-container failure locally (TH-managed OTBR container left running after an interrupted `start_device()` call blocked a subsequent run with `Border router does not start properly`); confirmed manually removing the orphaned container resolved it, which is what `remove_containers_for_image()` now automates.
- Have not reproduced the reporter's original `connect session failed` failure end-to-end; this is a hardening/diagnostics change plus a fix for a distinct, confirmed failure mode (orphaned containers) found during the investigation.

## Related

project-chip/certification-tool#1071


app/container_manager/container_manager.py
```diff
@@ -76,6 +76,22 @@ def get_container(self, id_or_name: str) -> Optional[Container]:
...
+        (like an RCP serial device) and block a new container from starting."""
+        stale_containers = self.__client.containers.list(
+            all=True, filters={"ancestor": docker_image_tag}
+        )
+        for container in stale_containers:
+            logger.warning(
+                f"Removing leftover container '{container.name}' "
+                f"({container.short_id}) for image {docker_image_tag}."
+            )
+            container.remove(force=True)
```
oxesoft: `force=True` here implicitly SIGKILLs the container if it's still `running`, with no attempt at a graceful stop first. That directly contradicts this same PR's rationale for switching `otbr_stop.sh` from `docker kill` to `docker stop` ("a hard kill can leave the RCP in a state that otbr_start.sh's next run fails to attach to"). If that theory is right, this path — the one most likely to touch a container in a stuck/unknown state — reintroduces the exact failure mode being fixed elsewhere.

Also worth noting: `ThreadBorderRouter.destroy_device()`'s normal teardown path still calls `container_manager.destroy()`, which does `container.kill()` (hard kill) unconditionally — untouched by this PR. So the graceful-stop fix only lands in the manual `otbr_stop.sh` script, not in the harness's own automated shutdown path.

Suggestion: try `container.stop(timeout=...)` before falling back to `remove(force=True)`, mirroring the existing `docker stop` → `docker rm` sequence a few lines above in `otbr_start.sh`.

app/container_manager/container_manager.py
```diff
@@ -76,6 +76,22 @@ def get_container(self, id_or_name: str) -> Optional[Container]:
             logger.info(f"Did not find container by id or name: {id_or_name}.")
             return None
 
+    def remove_containers_for_image(self, docker_image_tag: str) -> None:
```
oxesoft, line 102: No unit test added for this method. `app/tests/container_manager/test_container_manager.py` has focused tests for every other public `ContainerManager` method — given this one does a destructive, filter-based force-remove, a test mocking `containers.list`/`container.remove` seems worth adding.

app/container_manager/container_manager.py
```diff
@@ -76,6 +76,22 @@ def get_container(self, id_or_name: str) -> Optional[Container]:
...
+        """Force-remove any (running or stopped) container created from the given
+        image. Used to clear out containers that a previous, no-longer-tracked
+        instance of this process left behind (e.g. after a crash or a failed
+        start_device() call), which would otherwise hold onto host resources
+        (like an RCP serial device) and block a new container from starting."""
+        stale_containers = self.__client.containers.list(
+            all=True, filters={"ancestor": docker_image_tag}
```
oxesoft, line 117: This removes *every* container host-wide matching this image tag, not just ones this process orphaned — and `docker_image_tag` can come from user/config (`config.otbr_docker_image`). Probably fine given this image is dedicated to OTBR in the intended deployment, but might be worth a one-line comment documenting that assumption, since it's a fairly wide blast radius for a force-remove.

test_collections/matter/scripts/OTBR/otbr_start.sh
```diff
@@ -52,6 +52,15 @@ print_script_step "Removing 'otbr-chip' container"
 docker stop otbr-chip > /dev/null 2>&1
 docker rm otbr-chip > /dev/null 2>&1
 
+# Also remove any other leftover container from the OTBR image (e.g. one started by
+# the Test Harness itself via otbr_manager.py, which doesn't use the 'otbr-chip' name).
+# A stale container like this can still be holding the RCP serial device, which makes
+# otbr-agent fail to attach in the container we're about to start (see #1071).
+STALE_OTBR_CONTAINERS=$(docker ps -aq --filter ancestor=$BR_IMAGE)
+if [ -n "$STALE_OTBR_CONTAINERS" ]; then
+	echo "$STALE_OTBR_CONTAINERS" | xargs -r docker rm -f > /dev/null 2>&1
```
oxesoft: This force-removes stale containers directly (`docker rm -f`) without attempting a graceful stop first — inconsistent with the existing lines right above (52-53: `docker stop otbr-chip` then `docker rm otbr-chip`), and with this PR's own stated reasoning for why `otbr_stop.sh` was switched to graceful `docker stop`. If a leftover container is still actively running otbr-agent, hard-killing it here could leave the RCP in the same bad state this PR is trying to avoid. Consider `docker stop` before `-f` remove, e.g. `... | xargs -r docker stop; ... | xargs -r docker rm -f`.

oxesoft (commented):
Review findings inline. Overall the diagnostics additions (RCP-present check, readiness poll, log dumps) look solid; flagging a few things on the orphan-cleanup side before merging.

app/container_manager/container_manager.py
rquidute: Good catch — fixed. `destroy()` now has a `graceful` param (default False, so other callers keep the existing kill behavior) that tries `container.stop()` first and only falls back to `kill()` if the stop itself raises. `remove_containers_for_image()` and `otbr_manager.py`'s `destroy_device()` (the harness's own routine teardown, not just `otbr_stop.sh`) both now call `destroy(..., graceful=True)`, so the graceful-stop fix lands on all the paths that touch a container in a potentially stuck/unknown state, not just the manual script.

app/container_manager/container_manager.py
rquidute: Added tests in `app/tests/container_manager/test_container_manager.py`: `test_destroy_kills_by_default`, `test_destroy_graceful_stops_instead_of_kill`, `test_destroy_graceful_falls_back_to_kill_on_stop_failure`, `test_remove_containers_for_image_removes_stale_containers`, `test_remove_containers_for_image_no_stale_containers`.

app/container_manager/container_manager.py
rquidute: Fair point — added a note in the docstring calling out that this matches any container for the image host-wide, and that it's an accepted trade-off since these OTBR images are dedicated to TH use, not a general-purpose assumption.

test_collections/matter/scripts/OTBR/otbr_start.sh
rquidute: Fixed — now does `docker stop` on the stale containers before `docker rm -f`, matching the existing `otbr-chip` cleanup pattern right above it.

rquidute:
https://github.com/coderabbitai review full

rquidute:
@coderabbitai review full

test_collections/matter/scripts/OTBR/otbr_start.sh
```diff
@@ -52,6 +52,16 @@ print_script_step "Removing 'otbr-chip' container"
...
 docker rm otbr-chip > /dev/null 2>&1
 
+# Also remove any other leftover container from the OTBR image (e.g. one started by
+# the Test Harness itself via otbr_manager.py, which doesn't use the 'otbr-chip' name).
+# A stale container like this can still be holding the RCP serial device, which makes
+# otbr-agent fail to attach in the container we're about to start (see #1071).
+STALE_OTBR_CONTAINERS=$(docker ps -aq --filter ancestor=$BR_IMAGE)
+if [ -n "$STALE_OTBR_CONTAINERS" ]; then
+	echo "$STALE_OTBR_CONTAINERS" | xargs -r docker stop > /dev/null 2>&1
+	echo "$STALE_OTBR_CONTAINERS" | xargs -r docker rm -f > /dev/null 2>&1
```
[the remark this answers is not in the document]
rquidute: Good catch, fixed — the ancestor-filtered stop/rm now use `sudo docker`, consistent with the rest of the script. The "also applies to 100-100, 131-133" note looks stale though: both of those lines already used `sudo docker logs otbr-chip` before this fix, so no change was needed there.

oxesoft (approved):

antonio-amjr (approved):

PR #349, opened 2026-08-14
[Fix] Reuse of Admin Storage To Skip Commissioning 

antonio-amjr:
Fixes: https://github.com/project-chip/certification-tool/issues/1070
---

## Description

  Fixes the `admin_storage.json` counter-staleness half of #1070.

  `admin_storage.json` was only ever snapshotted to the host right after a fresh commissioning. Any test suite run that reused a previous commissioning restored that same original snapshot — never one reflecting counters as they stood after
  later test activity. This caused the DUT to see message counters go backwards on reuse, which it correctly (and problematically, for us) flags as a replay.

  This PR makes the harness re-capture `admin_storage.json` from the container at the end of every suite run (suite `cleanup()`, right before the container is destroyed), in addition to the existing capture right after commissioning. The host
  snapshot now always reflects the most recently advanced counters, whether or not that run recommissioned.

  ### Changes
  - `utils.py`: new `capture_admin_storage_file()`, a public wrapper around the existing private copy helper.
  - `test_suite.py`: `PythonTestSuite.cleanup()` now calls it before `sdk_container.destroy()`, guarded against `matter_config` never being set, and wrapped so a capture failure can never block teardown or fail the run.
  - Unit tests covering the happy path, the skip-when-unset case, and the capture-fails-but-teardown-still-runs case.

  ## Validation against the reported repro (TC-ACE-1.6 / TC-SC-5.2)

  Manually re-tested this fix against the exact reproduction in the issue (run once, then reuse commissioning, then run again). Findings, in case they're useful to whoever picks this up next:

  - **With this fix applied, and using a side-loaded copy of `TC_ACE_1_6.py` patched to de-duplicate its `GroupcastTesting` event waits, all runs pass reliably** — fresh commissioning and reused commissioning alike.
  - **This fix alone, against the unmodified upstream `TC_ACE_1_6.py`/`TC_SC_5_2.py`, is not sufficient.** Those scripts have a separate, pre-existing bug: several steps call the plain `wait_for_event_report(...)` on the `GroupcastTesting` 
  event, which pops whatever event is next in the queue without checking it actually corresponds to the command that step just sent. Real-hardware multicast delivery can produce duplicate events for a single group command, and those duplicates 
  sit in the queue and get consumed by a *later*, unrelated step — producing intermittent, non-deterministic failures unrelated to this counter-staleness bug.

    This isn't speculation on our part — the script's own author already flags it, twice:

    > "NOTE: WARNING - Multicast is not always reliable when testing over Wi-Fi, and the test currently does not implement retries. If the test fails (likely due to timeouts on operations such as receiving events), Try the test again from 
  factory reset a few times." (`TC_ACE_1_6.py`, module header)

    > "Duplicate events could occur from step 20d, as this is an event emitted from a point where the message cannot be decrypted (because of no group keys being present)... duplicate groupcast testing events can occur in certain cases (i.e. 
  when testing over multiple networks). This safely filters through potential duplicate events from the previous steps." (`TC_ACE_1_6.py`, step 20g)

    Step 20g has a working fix for exactly this (`wait_for_event_report_with_duplication`) — it's just not applied to the other affected call sites in either script. Full write-up with logs and the exact patch tested is attached to the issue.

  **This PR is the complete fix for what's addressable in this repo.** The event-duplication issue lives in the upstream SDK's test scripts (`connectedhomeip/src/python_testing`), not here, and per the header comment above, the SDK authors 
  already know retries are sometimes needed on real Wi-Fi hardware — there's nothing further to change on the certification-tool-backend side for that half of the problem.

  A generalized version of the dedup fix (filter on "current = the result this step expects, previous = anything else," applied uniformly across all the affected `GroupcastTesting` waits in both scripts) was validated locally and resolves the 
  intermittent failures. It's included in the linked write-up as a suggestion for the SDK scripts team to evaluate — it's their call whether it fits the test's intent, not something we're asserting should be merged as-is.

Attaching my **custom** version of the TC_ACE_1.6 to be verified if needed:
[TC_ACE_1_6_custom.py](https://github.com/user-attachments/files/31075561/TC_ACE_1_6_custom.py)

### DIFF
--- a/test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
+++ b/test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 from enum import Enum
-from typing import Type, TypeVar
+from typing import Optional, Type, TypeVar
 
 from app.constants.shared_constants import DutPairingModeEnum
 from app.schemas.test_environment_config import ThreadAutoConfig
@@ -58,6 +58,7 @@ class PythonTestSuite(TestSuite):
     suite_name: str
     sdk_container: SDKContainer = SDKContainer(logger)
     border_router: ThreadBorderRouter = ThreadBorderRouter()
+    matter_config: Optional[TestEnvironmentConfigMatter] = None
 
     @classmethod
     def class_factory(
@@ -118,14 +119,16 @@ class PythonTestSuite(TestSuite):
     async def cleanup(self) -> None:
         logger.info("Suite Cleanup")
 
-        matter_config = getattr(self, "matter_config", None)
-        if matter_config is not None and self.sdk_container.is_running():
+        if self.matter_config is not None and self.sdk_container.is_running():
             try:
                 logger.info(
                     "Capturing latest admin_storage.json snapshot from container"
                 )
-                capture_admin_storage_file(matter_config, logger)
+                capture_admin_storage_file(self.matter_config, logger)
             except Exception as e:
+                # Deliberately broad Exception.
+                # The ideia is to never block container/border-router teardown below,
+                # so don't narrow this to specific exception types.
                 logger.warning(f"Could not capture admin_storage.json snapshot: {e}")
 
         logger.info("Stopping SDK container")
@@ -138,6 +141,7 @@ class PythonTestSuite(TestSuite):
 class CommissioningPythonTestSuite(PythonTestSuite, UserPromptSupport):
     async def setup(self) -> None:
         await super().setup()
+        assert self.matter_config is not None
 
         # If in BLE-Thread, NFC-Thread, or THREAD_MESHCOP mode and a Thread Auto-Config
         # was provided by the user, start a new OTBR container app with the according
--- a/test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
+++ b/test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
@@ -30,6 +30,7 @@ from ...sdk_container import SDKContainer
 from ...utils import PromptOption, prompt_for_commissioning_mode
 from .utils import (
     DUTCommissioningError,
+    capture_admin_storage_file,
     commission_device,
     should_perform_new_commissioning,
 )
@@ -116,6 +117,16 @@ class PythonTestSuite(TestSuite):
 
     async def cleanup(self) -> None:
         logger.info("Suite Cleanup")
+
+        matter_config = getattr(self, "matter_config", None)
+        if matter_config is not None and self.sdk_container.is_running():
+            try:
+                logger.info(
+                    "Capturing latest admin_storage.json snapshot from container"
+                )
+                capture_admin_storage_file(matter_config, logger)
+            except Exception as e:
+                logger.warning(f"Could not capture admin_storage.json snapshot: {e}")
 
         logger.info("Stopping SDK container")
         self.sdk_container.destroy()
--- a/test_collections/matter/sdk_tests/support/python_testing/models/utils.py
+++ b/test_collections/matter/sdk_tests/support/python_testing/models/utils.py
@@ -207,6 +207,17 @@ def __copy_admin_storage_file(
     )
 
 
+def capture_admin_storage_file(
+    config: TestEnvironmentConfigMatter,
+    logger: loguru.Logger,
+) -> None:
+    """Re-capture admin_storage.json from the still-running container to the host
+    snapshot, so the snapshot reflects the message counters as they stood at the end
+    of this run rather than only as they stood right after the last commissioning.
+    """
+    __copy_admin_storage_file(config, logger)
+
+
 def log_test_output_file(logger: loguru.Logger) -> None:
     """Log the entire content of test_output.txt file.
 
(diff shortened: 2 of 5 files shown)


antonio-amjr:
@coderabbit full review

test_collections/matter/sdk_tests/support/python_testing/models/utils.py
```diff
@@ -207,6 +207,17 @@ def __copy_admin_storage_file(
...
 
+def capture_admin_storage_file(
+    config: TestEnvironmentConfigMatter,
+    logger: loguru.Logger,
+) -> None:
+    """Re-capture admin_storage.json from the still-running container to the host
+    snapshot, so the snapshot reflects the message counters as they stood at the end
+    of this run rather than only as they stood right after the last commissioning.
+    """
+    __copy_admin_storage_file(config, logger)
```
oxesoft, line 218: Now that `PythonTestSuite.cleanup()` unconditionally re-captures `admin_storage.json` at the end of every suite run, the existing capture right after commissioning (the other `__copy_admin_storage_file(config, logger)` call, in `commission_device` further down this file) looks redundant — every commissioning-based run will now do the docker-cp-equivalent copy twice (once there, once in `cleanup()`). Was that call intentionally kept as a fallback in case `cleanup()` is skipped on some path, or could it be dropped now that `cleanup()` covers it unconditionally?

test_collections/matter/sdk_tests/support/tests/python_tests/test_python_test_suite.py
```diff
@@ -444,6 +444,97 @@ async def test_should_perform_new_commissioning_yes() -> None:
...
+    ) as mock_capture:
+        await suite_instance.cleanup()
+
+    mock_capture.assert_not_called()
+    mock_destroy.assert_called_once()
+    mock_destroy_device.assert_called_once()
+
+
+@pytest.mark.asyncio
+async def test_cleanup_still_destroys_container_when_capture_fails() -> None:
```
oxesoft, line 540: The three new tests cover the happy path, `matter_config` unset, and capture raising — nice. One branch of the compound guard in `cleanup()` (`matter_config is not None and self.sdk_container.is_running()`) isn't covered yet: `matter_config` set but `is_running()` returning `False` (e.g. container died mid-suite). Worth a fourth test asserting `capture_admin_storage_file` is *not* called in that case, to lock in that the capture is skipped rather than attempted against a dead container.

test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
```diff
@@ -117,6 +118,16 @@ async def setup(self) -> None:
     async def cleanup(self) -> None:
         logger.info("Suite Cleanup")
 
+        matter_config = getattr(self, "matter_config", None)
```
oxesoft: `matter_config` is only ever assigned dynamically in `setup()` — it isn't declared as a class attribute the way `sdk_container`/`border_router` are above. That's why the `getattr(..., None)` fallback is needed here. Might be cleaner to declare `matter_config: Optional[TestEnvironmentConfigMatter] = None` on the class and just check `self.matter_config is not None` — same safety, but gives type-checkers something to work with and drops the string-literal attribute lookup.

test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
```diff
@@ -117,6 +118,16 @@ async def setup(self) -> None:
...
+        if matter_config is not None and self.sdk_container.is_running():
+            try:
+                logger.info(
+                    "Capturing latest admin_storage.json snapshot from container"
+                )
+                capture_admin_storage_file(matter_config, logger)
+            except Exception as e:
```
oxesoft, line 128: Catching bare `Exception` here is presumably deliberate — the PR description calls out that a capture failure must never block teardown — but it'll also silently downgrade unrelated bugs (e.g. a future `AttributeError` from a refactor) to a warning log instead of surfacing them. Might be worth a one-line comment here stating that's intentional, so a future pass doesn't narrow it and accidentally reintroduce teardown-blocking failures.

oxesoft (commented):
A few non-blocking points from review — the fix itself looks correct and is reasonably well tested.

rquidute (approved):

test_collections/matter/sdk_tests/support/python_testing/models/utils.py
antonio-amjr: It's intentionally redundant with cleanup()'s capture, kept as a fallback for abnormal teardown.

test_collections/matter/sdk_tests/support/tests/python_tests/test_python_test_suite.py
antonio-amjr: Done in: https://github.com/project-chip/certification-tool-backend/pull/349/commits/697401892c0d70126cca2daefa717ae91649562f

test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
antonio-amjr: Done in: https://github.com/project-chip/certification-tool-backend/pull/349/commits/697401892c0d70126cca2daefa717ae91649562f

test_collections/matter/sdk_tests/support/python_testing/models/test_suite.py
antonio-amjr: Deliberately broad: a failed capture is a "nice to have for the next run" convenience, not part of this run's pass/fail contract. It must never block container/border-router teardown below, so don't narrow this to specific exception types.

PR #354, opened 2026-08-21
Certification Mode - Fix pre-selection of test cases absent from the uploaded PICS

raul-marquez-csa:
#### Summary

Fixes an issue where Certification Mode pre-selected, and therefore forced the execution of, test cases belonging to clusters with no presence in the project's uploaded PICS (test cases whose scripts declare no PICS were treated as always applicable, in every collection), by auto-including test cases without declared PICS only when they belong to a mandatory collection or are explicitly named in the platform certification test list.

#### The issue

The applicable-test-cases computation carries a rule from the first TH release: a test case that declares no PICS is always applicable. At the time, the only tests without PICS were universal protocol tests, and this rule was the only way to keep them selected. Two things changed since:

- Mandatory collections became the proper mechanism for always-run tests (PICS v2, #122), but the old rule was kept in the non-mandatory pass as well.
- The SDK Python test collection grew many cluster tests whose scripts simply do not implement `pics_` (about 55 script files at the current SDK pin), and each of them inherits always-applicable status.

As a result, Certification Mode, whose selection is locked to the applicable set, forces test cases such as TC-SWTCH-2.x or TC-SEAR-1.x into runs for devices whose PICS never mention those clusters. Per the TH User Guide, in Certification Mode "a test case must be executed if and only if the PICS files indicate that it is applicable". Note this change also means that any non-mandatory test case without declared PICS is no longer pre-selected in either mode.

#### Included in this PR

- Auto-include test cases without declared PICS only for mandatory collections or when explicitly listed in the platform certification test list (`app/pics_applicable_test_cases.py`)
- Added unit tests covering mandatory and non-mandatory test cases without declared PICS, plus a combined scenario asserting that only uploaded-cluster tests and mandatory tests are selected (`app/tests/pics/test_applicable_test_cases_list.py`)

#### Testing

Applicability unit tests (18 passed; the two new negative tests fail against the previous selection logic)
```
python -m pytest app/tests/pics/test_applicable_test_cases_list.py
```
PR to fix CI
[CI - Fix spellcheck and Flake8 failures affecting all pull requests- #355](https://github.com/project-chip/certification-tool-backend/pull/355)

### DIFF
--- a/app/pics_applicable_test_cases.py
+++ b/app/pics_applicable_test_cases.py
@@ -184,8 +184,10 @@ def __applicable_test_cases(
                         continue
 
                     if not test_case.pics:
-                        # Test cases without PICS are always applicable
-                        applicable_tests.add(test_case.metadata["title"])
+                        # Test cases without PICS are only applicable if
+                        # mandatory or explicitly listed in tests_to_consider
+                        if test_collection.mandatory or tests_to_consider:
+                            applicable_tests.add(test_case.metadata["title"])
                     else:
                         test_enabled_pics, test_disabled_pics = __retrieve_pics(
                             test_case
--- a/app/tests/pics/test_applicable_test_cases_list.py
+++ b/app/tests/pics/test_applicable_test_cases_list.py
@@ -330,6 +330,127 @@ def test_applicable_test_cases_set_with_mandatory_tests(mock_manager) -> None:
 
 
 @patch("app.pics_applicable_test_cases.test_script_manager")
+def test_applicable_test_cases_set_includes_mandatory_tests_without_pics(
+    mock_manager,
+) -> None:
+    # Create PICS without any PICs relevant to the test case
+    pics = create_random_pics()
+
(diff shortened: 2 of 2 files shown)


KaHa6uc:
Thanks for this — we hit exactly the same issue on a real DUT, and TC-SWTCH-2.x is our example too.

For context: we're certifying a Matter device on this harness (backend `2ac562d`, SDK `db418e8e`). `TC_SwitchTests` exposes `test_TC_SWTCH_2_2` .. `2_6`, all with `pics: []`, and all five were force-selected onto a product that implements no Generic Switch. They failed on the missing `--endpoint` flag rather than skipping, so we carry a local `PER_TEST_PARAMETERS` workaround whose only purpose is to turn those false FAILEDs into NOT_APPLICABLE. _So: strongly in favour of the direction._

cecille (approved):

rquidute (approved):

oxesoft (approved):

antonio-amjr (approved):

PR #356, opened 2026-08-21
[Feature] Adding THconfig schema as a new TH config field

### LINKED ISSUE #1038: Configurable timeout for user prompts (currently hardcoded to 60 seconds)
## Description

User prompts shown during test execution currently time out after a fixed 60 seconds, and this value cannot be configured.

## Motivation

Some test steps require the operator to manually read through test output and take an action in response to a prompt — a task that can take longer than 60 seconds, causing the prompt to time out before the operator can respond.

This was flagged as related to an existing bug report, but is being tracked separately since it's a general feature request (configurable prompt timeout) rather than an issue specific to one test case.

## Proposed change

Make the user-prompt timeout configurable instead of hardcoded, so it can be increased where operators need more time to respond.

## Related
- #973

antonio-amjr:
Fix: https://github.com/project-chip/certification-tool/issues/1038
Related to:

1. https://github.com/project-chip/matter-test-scripts/pull/801
2. https://github.com/project-chip/certification-tool-cli/pull/109
---

## Description
User prompts during test execution timed out after a hardcoded 60s with no way to configure it. Adds a th_config object to the project's JSON config (alongside network/dut_config/test_parameters) holding Test-Harness-level
  settings, starting with prompt_timeout_seconds. Also folds the existing ENABLE_REALTIME_PYTHON_TEST_LOGS env var into th_config as a per-project override, with the env var kept as the instance-wide fallback (non-breaking).

  Changed
  - schemas/test_environment_config.py: new THConfig (prompt_timeout_seconds: int = 60, enable_realtime_python_test_logs: Optional[bool] = None), added as TestEnvironmentConfig.th_config. No DB migration — config is a schemaless JSON column.
  - user_prompt_support/prompt_request.py: PromptRequest.timeout is now Optional[int] = None (unresolved sentinel).
  - user_prompt_support/user_prompt_support.py: send_prompt_request() resolves an unresolved timeout from self.config["th_config"]["prompt_timeout_seconds"], falling back to 60s. Explicit per-call timeouts (120s/300s/600s tiers elsewhere) are
  untouched.
  - user_prompt_support/user_prompt_manager.py: defensive fallback in PromptExchange in case a future call site bypasses the funnel above.
  - Cleaned up 5 remaining redundant hardcoded timeout=60 sites (chip_suite.py, chip_test.py) and one dead constant.
  - python_testing/models/test_case.py: new _realtime_logs_enabled() — th_config.enable_realtime_python_test_logs overrides settings.ENABLE_REALTIME_PYTHON_TEST_LOGS when explicitly set.
  - default_project.config: added example th_config block.
  
  Test plan
  - New unit tests: timeout resolution (valid/missing/malformed config, explicit-override-wins), THConfig schema validation, real-time-log toggle matrix (config × env var).
  - Existing prompt-support and Matter config tests still pass — verified no test asserts a literal timeout == 60.
  - Not yet run in CI by me (no local backend env) — please run pytest app/tests/user_prompts/ test_collections/matter/sdk_tests/support/tests/ before merging.

### DIFF
--- a/app/schemas/test_environment_config.py
+++ b/app/schemas/test_environment_config.py
@@ -48,11 +48,21 @@ class TestEnvironmentConfigError(Exception):
     """
 
 
+class THConfig(BaseModel):
+    """Test-Harness-specific settings, independent of any particular DUT/program."""
+
+    prompt_timeout_seconds: int = 60
+    # None means "not set at project level", deferring to the
+    # ENABLE_REALTIME_PYTHON_TEST_LOGS environment variable.
+    enable_realtime_python_test_logs: Optional[bool] = None
+
+
 class TestEnvironmentConfig(BaseModel):
     __test__ = False  # Needed to indicate to PyTest that this is not a "test"
 
     # TODO(#490): Need to be refactored to support real PIXIT format
     test_parameters: Optional[dict[str, Any]]
+    th_config: Optional[THConfig] = None
 
     def __init__(self, **kwargs: Any):
         try:
--- a/app/tests/user_prompts/test_options_select_user_prompts.py
+++ b/app/tests/user_prompts/test_options_select_user_prompts.py
@@ -193,6 +193,7 @@ async def test_options_select_user_prompt_manager_prompt(db: Session) -> None:
         # Assert broadcast was first called twice with:
         # 1. prompt_dict
         # 2. timeout_notification
+        assert prompt.timeout is not None
         expected_broadcast_calls = [
             call(__expected_prompt_dict(timeout=prompt.timeout)),
             call(__expected_timeout_notification_dict()),
--- a/app/tests/user_prompts/test_prompt_exchange.py
+++ b/app/tests/user_prompts/test_prompt_exchange.py
@@ -21,9 +21,32 @@ import pytest
 
 from app.user_prompt_support import user_prompt_manager
 from app.user_prompt_support.constants import UserResponseStatusEnum
-from app.user_prompt_support.prompt_request import PromptRequest
+from app.user_prompt_support.prompt_request import PromptRequest, default_timeout_s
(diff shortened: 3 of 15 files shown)


app/schemas/test_environment_config.py
```diff
@@ -48,11 +48,21 @@ class TestEnvironmentConfigError(Exception):
...
 
+class THConfig(BaseModel):
+    """Test-Harness-specific settings, independent of any particular DUT/program."""
+
+    prompt_timeout_seconds: int = 60
+    # None means "not set at project level", deferring to the
+    # ENABLE_REALTIME_PYTHON_TEST_LOGS environment variable.
+    enable_realtime_python_test_logs: Optional[bool] = None
+
+
```
rquidute, line 59: Please also add `enable_container_logs`. Refer to user guide `12.5.2.4. Python Test Logging Configuration`

rquidute (changes requested):

rquidute (approved):

oxesoft (approved):

app/schemas/test_environment_config.py
antonio-amjr: Hey @rquidute,

I would recommend to postpone this request to a new Issue. The real-time log env variable was a low risk straight forward addition but `enable_container_logs` seems not. The enable container variable will require adding parameters through **Singleton** infrastructure (`ContainerManager`, `SDKContainer` and more) across several calls.
Another reason would be the already complex review already in hand for this change (20+ files between 4 repos), besides the unit tests required to include for those.

That said, is it ok to hold this for later? Let me know if you agree

rquidute: Agreed @antonio-amjr

PR #359, opened 2026-08-26
[CP] Massive Logs Support to CLI (#347) to v2.16-develop

rquidute:
## Summary

Cherry-picks 35f877f ("[Fix] Massive Logs Support to CLI", originally merged as #347) from `v2.15.1-develop` onto `v2.16-develop`, which never received it.

## Why

While investigating a report that test executions "run fast but get stuck at the last step, seemingly processing a huge number of log lines," I traced it to four compounding issues in the log pipeline that are exactly what #347 already fixed upstream:

1. **Blocking, redundant DB commits** (`test_db_observer.py`): every ~0.5s log-flush tick re-queued the *same* `test_run_execution` object (with its ever-growing `.log` list) onto a plain `Queue`. At the end of the run, draining that queue ran a **synchronous `session.commit()` on the event loop once per queued item**, redundantly re-saving the same, increasingly large log blob over and over — this is the "stuck at the end" symptom.
2. **Unchunked websocket broadcasts** (`test_ui_observer.py`): a burst of log lines was sent as one multi-MB websocket message with no yield point during JSON serialization, which can stall the event loop long enough to miss keepalive pongs.
3. **Missed websocket keepalives under load** (`main.py`): uvicorn's default 20s `ws_ping_timeout` wasn't overridden, so the stall from #2 could get the websocket dropped mid-broadcast.
4. **Whole-file re-reads for SDK step logs** (`test_case.py`): the incremental log tailer had encoding/step-tracking issues that could reprocess large chunks of `test_output.txt` repeatedly.

`v2.16-develop` is missing this fix entirely (confirmed via `git merge-base --is-ancestor 35f877f` returning false), while `v2.16-develop` is what recent feature work (e.g. #357) is based on — so this regression is currently live on that line.

## Conflict resolution

One cherry-pick conflict in `test_collections/matter/sdk_tests/support/tests/python_tests/test_python_test_case.py`: both branches independently appended unrelated test functions at the same insertion point. Resolved by keeping both sides' additions in full — verified via AST comparison that every function from both parent commits is present in the merged file with an identical AST to its original (i.e. no test was dropped, corrupted, or duplicated), then ran `black` to normalize the seam's blank-line spacing.

## Testing

- `black`, `isort`, `flake8`, `mypy` all pass on every changed file (mypy's 1 remaining finding — `test_db_observer.py`'s `insp.session` attr-defined error — is pre-existing on `v2.16-develop` prior to this cherry-pick, confirmed by running mypy against the file before applying this PR).
- Verified via AST-level diff that the conflict resolution preserved 100% of both parents' test functions (32 = 16 base + 7 added upstream in this fix's own tests + 9 added independently on `v2.16-develop` since diverging).
- Full `pytest` suite (Docker/DB-dependent) was not run in the authoring environment; please run `./scripts/test-local.sh` before merge.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -20,6 +20,7 @@ from starlette.middleware.cors import CORSMiddleware
 from app.api.api_v1.api import api_router
 from app.core.config import settings
 from app.test_engine.test_script_manager import test_script_manager
+from app.uvicorn_worker import WS_PING_TIMEOUT_S
 
 app = FastAPI(
     title=settings.PROJECT_NAME, openapi_url=f"{settings.API_V1_STR}/openapi.json"
@@ -51,4 +52,9 @@ async def startup_event() -> None:
 
 
 if __name__ == "__main__":
-    uvicorn.run(app, host="0.0.0.0", port=80, log_config=None)
+    # ws_ping_timeout raised from uvicorn's 20s default: a large log flush can
+    # keep the event loop busy long enough to miss the keepalive pong and get
+    # the websocket dropped mid-broadcast (see app.uvicorn_worker).
+    uvicorn.run(
+        app, host="0.0.0.0", port=80, log_config=None, ws_ping_timeout=WS_PING_TIMEOUT_S
+    )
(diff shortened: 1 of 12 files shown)


antonio-amjr (approved):

antonio-amjr:
Gentle reminder: please cherry-pick also the [CLI_PR#107](https://github.com/project-chip/certification-tool-cli/pull/107) to the CLI repo

oxesoft (approved):

PR #368, opened 2026-09-09
Add wpa_supplicant mount for CNET tests in SDKContainer

khodya:
## Summary

  Adds a `/var/run/wpa_supplicant` volume mount to the `chip-cert-bins` (SDK)
  container that the Test Harness starts automatically, so CNET (Network
  Commissioning cluster) test cases can manage Wi-Fi networks on the host
  without requiring testers to start the container manually with an extra `-v`
  flag.

  ## Background

  CNET Wi-Fi network commissioning tests talk to the host's `wpa_supplicant`
  over its control socket. `SDKContainer.run_parameters["volumes"]` (used by
  `container_manager.create_container(...)` when the TH engine/UI boots the
  container) already mounts `/var/run/dbus/system_bus_socket`, PAA certs,
  credentials, and the python_testing tree — but not `/var/run/wpa_supplicant`.
  Without it, CNET tests run through the automated TH engine can't reach
  `wpa_supplicant`, forcing manual `docker run ... -v 
  /var/run/wpa_supplicant:/var/run/wpa_supplicant ...` workarounds outside the 
  TH.

  ## Changes

  - `test_collections/matter/sdk_tests/support/sdk_container.py`:
    - Added `LOCAL_WPA_SUPPLICANT_PATH` / `DOCKER_WPA_SUPPLICANT_PATH` constants
  (`/var/run/wpa_supplicant`), following the existing mount-constant pattern.
    - Added the corresponding entry to `SDKContainer.run_parameters["volumes"]`
  (`mode: rw`, matching the D-Bus socket mount).

  ## Test plan

  - [ ] Start the SDK container via the TH engine and confirm
  `/var/run/wpa_supplicant` is bind-mounted (`docker inspect <container> | grep
  wpa_supplicant`).
  - [ ] Run a CNET test case through the TH and confirm it can add/update/remove
  Wi-Fi networks without the manual volume workaround.

### DIFF
--- a/test_collections/matter/sdk_tests/support/sdk_container.py
+++ b/test_collections/matter/sdk_tests/support/sdk_container.py
@@ -47,6 +47,10 @@ DOCKER_PAA_CERTS_PATH = "/paa-root-certs"
 LOCAL_CREDENTIALS_DEVELOPMENT_PATH = Path("/var/credentials/development")
 DOCKER_CREDENTIALS_DEVELOPMENT_PATH = "/credentials/development"
 
+# wpa_supplicant mount (required for CNET tests to manage Wi-Fi networks)
+LOCAL_WPA_SUPPLICANT_PATH = Path("/var/run/wpa_supplicant")
+DOCKER_WPA_SUPPLICANT_PATH = "/var/run/wpa_supplicant"
+
 # Python Testing Folder
 LOCAL_TEST_COLLECTIONS_PATH = (
     "/home/ubuntu/certification-tool/backend/test_collections/matter"
@@ -106,6 +110,10 @@ class SDKContainer(metaclass=Singleton):
             LOCAL_CREDENTIALS_DEVELOPMENT_PATH: {
                 "bind": DOCKER_CREDENTIALS_DEVELOPMENT_PATH,
                 "mode": "ro",
+            },
+            LOCAL_WPA_SUPPLICANT_PATH: {
+                "bind": DOCKER_WPA_SUPPLICANT_PATH,
+                "mode": "rw",
             },
             LOCAL_PYTHON_TESTING_PATH: {
                 "bind": DOCKER_PYTHON_TESTING_PATH,


greens (commented):
Not exactly my area of expertise, but would this still work on a host that's not running supplicant or has no wifi interfaces?

cecille:
Are there pi's that don't have wifi interfaces?

cecille (approved):
We've discussed this change already offline. I'm not a docker expert, but this looks correct to me.

greens:
@cecille unsure, but would this break, for example, any developer attempting to run locally on a desktop machine, server, or mini-pc without one?

I understand the Pi is the target platform, but I think broad compatibility is still a good goal.

oxesoft:
@khodya In case this is expected for 1.7 TE2, please, change the target branch to `v2.16-develop`.

oxesoft:
This mount is added unconditionally to the class-level `run_parameters["volumes"]` dict, so it applies to every `SDKContainer.start()` call — i.e. every test run, not just CNET/Wi-Fi tests. On hosts without `wpa_supplicant` (no Wi-Fi hardware, wired-only mini-PCs/servers, or non-Linux Docker hosts), this could break SDK container creation entirely for all tests, not just CNET ones:

- Docker Desktop for Mac only shares `/Users`, `/Volumes`, `/private`, `/tmp`, `/var/folders` by default — `/var/run` isn't on that list, so binding it would likely fail with "Mounts denied."
- Rootless Docker on Linux runs the daemon unprivileged and typically can't auto-create a directory under root-owned `/var/run`.

Suggest gating the mount on the host path actually existing, checked at container-start time rather than baked into the static class attribute, e.g.:

```python
def _optional_volumes() -> dict:
    volumes = {}
    if LOCAL_WPA_SUPPLICANT_PATH.is_dir():
        volumes[LOCAL_WPA_SUPPLICANT_PATH] = {
            "bind": DOCKER_WPA_SUPPLICANT_PATH,
            "mode": "rw",
        }
    return volumes
```

and merge it into `volumes` inside `start()` before calling `container_manager.create_container(...)`. That way hosts without `wpa_supplicant` behave exactly as they did before this PR (CNET Wi-Fi tests just don't work, same as requiring the manual `-v` flag today), instead of risking a failure to start the SDK container for every test.

oxesoft:
@khodya Could you please address the comment above ASAP? We are about to kick off 1.7 TE2 and the related documentation was [already integrated](https://github.com/project-chip/certification-tool/pull/1108).

khodya:
@oxesoft fixed, thanks!

greens (approved):

oxesoft (approved):

rquidute (approved):

PR #369, opened 2026-09-11
Restrict avahi allow-interfaces to eth0, wpan0, wlan0

khodya:
## Summary
- The OTBR container's `avahi-daemon.conf` had `allow-interfaces=` (empty), which allows avahi to bind to *all* interfaces on the host, including Docker's virtual bridge/veth interfaces.
- On hosts with several Docker networks, this let stale mDNS records persist on virtual interfaces, which broke Thread device discovery during test runs.
- Restrict `allow-interfaces` to the interfaces actually relevant to OTBR: `eth0` (host uplink), `wpan0` (Thread radio interface), `wlan0` (Wi-Fi, used in some test setups).
- This is required for CNET-4.12 (network switching test) to pass reliably: the test switches the DUT between Thread networks, and OTBR acts as one of the Thread networks it switches to. Stale mDNS records on Docker virtual interfaces caused discovery of the OTBR-backed network to fail intermittently during that switch.

## Test plan
- [ ] Run CNET-4.12 and confirm the DUT successfully switches to and discovers the OTBR-backed Thread network.
- [ ] Run other OTBR-backed Thread test cases and confirm discovery no longer picks up stale mDNS records from Docker virtual interfaces.
- [ ] Verify OTBR container still advertises/discovers correctly over `eth0`/`wpan0`/`wlan0` on a test harness with active Docker networks.


test_collections/matter/sdk_tests/support/otbr_manager/avahi/avahi-daemon.conf
```diff
@@ -39,7 +39,7 @@ host-name=otbr
 #browse-domains=0pointer.de, zeroconf.org
 use-ipv4=no
 use-ipv6=no
-allow-interfaces=
+allow-interfaces=eth0,wpan0,wlan0
```
cecille, line 42: do you know that these interfaces are the only ones that are going to be present?

Would the disallow list be better here (ie, disallow docker / local?)

cecille (approved):

antonio-amjr (approved):

test_collections/matter/sdk_tests/support/otbr_manager/avahi/avahi-daemon.conf
khodya: it is hard to use a deny list as well, since docker veth interfaces get random names: e.g. `vethXXXXXXX` and there is no a wildcard syntax

